PR #3, opened 2018-05-31
Implement Footer component

sarathms:
A static footer with links. Based on the UI mockup.

### DIFF
--- a/components/footer.js
+++ b/components/footer.js
@@ -12,8 +12,8 @@ const StyledFooter = styled.footer`
 `
 
 const FooterBox = styled(Box)`
-  padding-top: 50px;
-  padding-bottom: 50px;
+  padding-top: 25px;
+  padding-bottom: 25px;
 `
 
 const FooterHead = styled.div`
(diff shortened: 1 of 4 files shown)


hellais:
Great job on the footer! It's perfect. We will perhaps at some point integrate in the ooni-components design system for maximum re-use across all our react based web interfaces.

hellais:
The only thing I would suggest you do is:
- Add the Footer component to the Layout component so that it appears in every page (maybe add an optional prop to disable it via `<Layout disableFooter={true}>`
- Make sure it work properly with pages that don't have enough content on top (i.e. sticky)

sarathms:
> We will perhaps at some point integrate in the ooni-components design system for maximum re-use across all our react based web interfaces.

As we finish building components here, we can move some of them to ooni-components if/when we see possible abstractions. Is that what you also meant?

> Add the Footer component to the Layout component so that it appears in every page (maybe add an optional prop to disable it via <Layout disableFooter={true}>

Good idea.

hellais:
> As we finish building components here, we can move some of them to ooni-components if/when we see possible abstractions. Is that what you also meant?

yes exactly.

> Good idea.

👍

PR #5, opened 2018-06-05
Feature/ci

hellais:


scripts/now-preview.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -ex
+
+now --token ${NOW_TOKEN} rm -y ooni-explorer || echo "nothing deleted"
+NOW_URL=$(now --token ${NOW_TOKEN} ./ -n ooni-explorer --public --npm)
```
sarathms: Doesn't this mean that there will be only one instance running at any given time?

hellais: Correct. The reason to do that, though, is that the now free plan only covers up to 3 instances (which is a pretty tight limit), so I figured it was better to have at least one running for sure, than risking to fail to make a build.

PR #16, opened 2018-09-04
Feature/measurements

hellais:


components/flag.js
```diff
@@ -25,31 +25,41 @@ var supportedCountryCodes = [
...
-      <FlagImg src={src} />
+    <FlagContainer className='country-flag' size={size}>
+      <FlagImg src={src} size={size} />
     </FlagContainer>
   )
 }
 
+Flag.propTypes = {
+  countryCode: PropTypes.string.isRequired,
+  size: PropTypes.number,
```
hellais: This should probably also be an `isRequired`

components/measurement/CommonDetails.js
```diff
@@ -0,0 +1,124 @@
...
+const softwareNames = {
+  'ooniprobe-ios': {
+    software: 'OONI Probe',
+    platform: 'iOS'
+  },
+  'ooniprobe-android': {
+    software: 'OONI Probe',
+    platform: 'Android'
+  }
+}
```
hellais: I think it's more robust to get the target platform from the `annotations->platform` key. This way we can get it using the same strategy for `ios`, `macos`, `android`, `windows`, `linux` and `lepidopter`.

hellais: See the base data format spec WRT to the `annotations` key: https://github.com/ooni/spec/blob/master/data-formats/df-000-base.md#specification

components/measurement/CommonDetails.js
```diff
@@ -0,0 +1,124 @@
...
+            <JsonViewer src={measurement} />
+          </NoSSR>
+        </Flex>
+      </Box>
+    </Container>
+  )
+}
+
+CommonDetails.propTypes = {
+  measurement: PropTypes.object.isRequired
```
hellais: I wonder if we should be encapsulating the knowledge about how the measurement is structured in this component or if it should be the responsibility of the parent to pass in only the props that are needed to show this view.

I think this is probably a good approach since we actually need the whole measurement for the JsonViewer, but if we notice that there are some performance issues we may actually add support for fetching only a subset of the keys in the API and fetch the whole measurement only when the JsonViewer is opened.

components/measurement/CommonSummary.js
```diff
@@ -0,0 +1,145 @@
...
+import MdPhonelink from 'react-icons/lib/md/phonelink'
+import MdRestore from 'react-icons/lib/md/restore'
+
+import NavBar from '../nav-bar'
+import Flag from '../flag'
+import Badge from '../badge'
+import { testGroups, testNames } from '../test-info'
+
+const SummaryContainer = styled.div`
+  background-color: ${props => props.color };
```
hellais: nitpick: extra space after `props.color`.

pages/measurement.js
```diff
@@ -152,17 +21,24 @@ export default class Measurement extends React.Component {
...
-    let msmtResult = await client.get('/api/v1/measurements', {
-      params
-    })
-
+    let [msmtResult, countriesR] = await Promise.all([
+      client.get('/api/v1/measurements', {
+        params
+      }),
+      client.get('/api/_/countries')
+    ])
```
hellais: I don't remember what happens if one of the two promises fails and what is the correct pattern for handling this, but this is probably something to think about.

components/measurement/nettests/web_connectivity.js
```diff
@@ -65,6 +65,21 @@ const HttpResponseBodyContainer = styled(Pre)`
   overflow: auto;
 `
 
+const FailureString = ({failure}) => {
+  if (!failure) {
+    return (
+      <div>
+        <Tick size={20} /> null
```
sarathms: Unless the `null` is intentional, will remove it in as I go ahead.

hellais:
Yes the null is intentional as it’s to show how we represent no failure in the data

components/measurement/CommonSummary.js
sarathms: Fixed

components/flag.js
sarathms: :+1:

components/measurement/nettests/web_connectivity.js
```diff
@@ -0,0 +1,180 @@
...
+
+const StatusBar = ({
+  anomaly,
+  hint
+}) => (
+  <Box mb={4} p={3} color='white' bg={anomaly ? 'yellow8' : 'green7'}>
+    <Container>
+      <Flex>
+        <Box w={1/2}>
+          {anomaly ? <StatusLabelAnomaly /> : <StatusLabelOK />}
```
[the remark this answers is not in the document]
sarathms: The way it was done, `<StatusBar>` only needs to check if `anomaly` is truthy or not. But I think I diverged from why I decided to switch it from a boolean to a ENUM-ish structure.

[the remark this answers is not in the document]
sarathms: Since there are imminent changes on this component, will postpone this until we settle the design questions surrounding this one.

components/measurement/nettests/web_connectivity.js
```diff
@@ -0,0 +1,180 @@
...
+      anomaly = 'HTTPDIFF'
+      hint = 'Different HTTP Response'
+    } else if (blocking === 'http-failure') {
+      anomaly = 'HTTPFAILURE'
+      hint = 'HTTP Reqeust Failed'
+    } else if (blocking === 'tcp-ip') {
+      anomaly = 'TCPIP'
+      hint = 'TCP/IP Based Blocking'
+    }
+  }
```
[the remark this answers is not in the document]
sarathms: That was one of the directions I considered, but I thought rendering the rest of the details might also depend on the type of anomaly detected. So left it out in the open there.

components/measurement/CommonDetails.js
hellais: I guess the main thing we could optimise in terms of fetching are the HTTP response bodies, which we could also make toggle-able. For other measurements that are not the HTTP one, it's probably true that we don't gain too much from following this approach.

components/measurement/nettests/web_connectivity.js
hellais: Yes the `null` is intentional as it’s to show how we represent no failure in the data

pages/measurement.js
sarathms: There is a [this pattern](https://davidwalsh.name/promises-results) that I just came across. Maybe a utility function to keep it readable? Like this:

hellais: Actually I think we are better of failing if one of the promises fail as it will likely break the UI. Maybe we should do something with the event like log it to the error log.

gabelula:
@sarathms what is the situation with this PR and review?

sarathms:
@gabelula This branch has now been brought on par with `master` branch which underwent some dependency upgrades. Once I deploy this on the test instance (early tomorrow), I will merge this into master, close this PR and open a new PR to develop/review the implementation of latest mockups.

PR #17, opened 2018-09-18
Upgrade dependencies to new versions

sarathms:
Target dependencies: 
```
"rebass": "2.3.3",
"@rebass/grid": "6.0.0",
"styled-components": "4.0.0-beta.3",
"styled-system": "3.1.3"
```

### DIFF
--- a/components/globe.js
+++ b/components/globe.js
@@ -84,7 +84,7 @@ class WebGLGlobe extends React.Component {
 
     if (webGLEnabled) {
       return (
-        <StyledGlobe ref={(el) => { this.globeRef = el }} />
+        <StyledGlobe innerRef={(el) => { this.globeRef = el }} />
       )
     } else {
       return <div>Your browser does not support WebGL</div>
--- a/components/globe.js
+++ b/components/globe.js
@@ -84,7 +84,7 @@ class WebGLGlobe extends React.Component {
 
     if (webGLEnabled) {
       return (
-        <StyledGlobe innerRef={(el) => { this.globeRef = el }} />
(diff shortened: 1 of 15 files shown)


sarathms:
There might be a few more commits coming up after checking if `styled-component`'s API changes have more impact in places I haven't looked yet.

components/globe.js
```diff
@@ -84,7 +84,7 @@ class WebGLGlobe extends React.Component {
 
     if (webGLEnabled) {
       return (
-        <StyledGlobe innerRef={(el) => { this.globeRef = el }} />
+        <StyledGlobe ref={(el) => { this.globeRef = el }} />
```
hellais: It's cool that they have fixed this. Do you know if this requires a particular version of React (I seem to recall this wasn't possible in previous styled-components version due to react limitations)? 

Be sure to do some testing of this (especially when used in form/input components) as there may be some breakage.

components/layout.js
```diff
@@ -11,7 +11,7 @@ import Footer from './footer'
 
 theme.maxWidth = 1024
 
-injectGlobal`
+const GlobalStyle = createGlobalStyle`
```
hellais: We should probably add this to design-system

components/globe.js
sarathms: This is based on the `forwardRef` feature coming in with [`react@16.3.0`](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#forwardref-api). Which means we have to upgrade react too which seems to be a non-breaking upgrade. We should expedite writing tests.

Issue #38, opened 2018-12-21
Download measurement data in json 


gabelula:
The download measurement button uses Javascript and `datauri` to make the raw measurement be donwloadable. We should instead use `href` pointing to the raw measurement on the OONI API. 

This allows non-js browser to download it, overcomes the datauri filesize limitatio and also makes it easier to share the download link.

hellais:
This is done in: https://github.com/ooni/explorer/commit/263594d0dec96df5e129235a56d1190b68659c91

PR #42, opened 2019-01-01
Improvements to measurement pages

hellais:
These are mostly copy improvements and some bug fixes.

I also implement what is specified in: https://github.com/ooni/explorer/issues/38

### DIFF
--- a/components/measurement/CommonDetails.js
+++ b/components/measurement/CommonDetails.js
@@ -11,8 +11,6 @@ import {
 } from 'ooni-components'
 
 import NoSSR from 'react-no-ssr'
-import styled from 'styled-components'
-import jsFileDownload from 'js-file-download'
 
 import DetailsBox from './DetailsBox'
 
@@ -34,7 +32,8 @@ JsonViewer.propTypes = {
 }
 
 const CommonDetails = ({
-  measurement
+  measurement,
+  measurementURL
 }) => {
   const {
     report_id,
@@ -45,6 +44,7 @@ const CommonDetails = ({
     }
   } = measurement
 
+  const downloadFilename = `ooni-measurement-${report_id}.json`
   const items = [
     {
       label: 'Measurement ID',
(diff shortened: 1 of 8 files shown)


elioqoshi:
Who should ideally review this? @sarathms ?

hellais:
> Who should ideally review this? @sarathms ?

Yes.

components/measurement/CommonDetails.js
```diff
@@ -80,15 +80,12 @@ const CommonDetails = ({
...
-            <Button
-              onClick={() => (
-                jsFileDownload(
-                  JSON.stringify(measurement, null, 2),
-                  'ooni-measurement-' + report_id + '.json')
-              )}
-              fontSize={11}
-              mx={3}
-              px={3}>Download JSON</Button>
+            <a href={measurementURL} download={downloadFilename}>
```
sarathms: This seems to work as a download link only when relative URLs are used. "Save Link As..." works but doesn't pick up the `downloadFilename` value.

components/measurement/StatusHeader.js
```diff
@@ -18,25 +18,24 @@ const StatusHeader = ({ status, color, icon, label, info }) => {
...
-        icon = <MdCheck />
-        break
-      case 'confirmed':
-        computedLabel = 'Blocking Confirmed'
-        icon = <MdCheck />
-        break
-      default:
-        computedLabel: ''
-        icon: <div />
+    case 'anomaly':
```
sarathms: This needs to be checked again. [Eslint documentation](https://eslint.org/docs/rules/indent#switchcase) says the earlier format is the correct one. Of course, this depends on what format we choose to follow.

components/measurement/CommonDetails.js
hellais: I am going to check if some backend change is necessary in order to make it download instead of showing it in the browser. However I also don't mind it appearing in the browser and prefer that to a client-side JS solution.

components/measurement/StatusHeader.js
hellais: You mean the level of indentation?

My editor which should be using the eslint configuration (via `eslint --fix`) to autofix the fixable did that.

Running `yarn run lint` I see no errors in that particular file.

Do you use an editor plugin to auto-indent stuff and fix linting errors? Maybe we should just align our editor configs.

components/measurement/CommonDetails.js
hellais: I filed this API ticket for this: https://github.com/ooni/api/issues/68

hellais: This is implemented in: https://github.com/ooni/api/pull/69

components/measurement/StatusHeader.js
sarathms: I do use the same atom package. Looks like this is a side effect of some other tweak I made to my config elsewhere :man_facepalming:

PR #48, opened 2019-01-16
Replace localization strings with react-intl declarations

### LINKED ISSUE #46: Convert all copy strings into keys
Similarly to the mobile app copy we should make all the strings into keys and put them inside of a spreadsheet so that @agrabeli can write copy for them (and also support translating them).

See: https://docs.google.com/spreadsheets/d/1GPkr_OyNhXJRTXnseyNbo2P4Du-i_W9kJdqjZG4HVkM/edit#gid=0

sarathms:
Fixes #46

## Proposed Changes
  - Use `react-intl` features like `<FormattedMessage />` and `intl.formatMessage` to define keys
  - Extract those keys to a spreadsheet/csv which can be used for localization efforts.
  - Localized strings are compiled into json files used by `react-intl` to render pages


components/measurement/nettests/web_connectivity.js
```diff
@@ -242,25 +244,34 @@ const WebConnectivityDetails = ({ measurement, render }) => {
...
+              <DetailsBox pl={2} title={<FormattedMessage id='Measurement.Details.Websites.TCP.Heading' />} content={
                 <React.Fragment>
-                  {tcpConnections.length === 0 && <Text>No results</Text>}
+                  {tcpConnections.length === 0 && <Text><FormattedMessage id='Measurement.Details.Websites.TCP.NoData' /></Text>}
                   {tcpConnections.map((connection, index) => (
                     <Flex key={index}>
                       <Box>
-                        <Text>Connection to <strong>{connection.destination}</strong> was {connection.status}.</Text>
+                        <Text>
+                          <FormattedMessage
```
sarathms: This one has parameters and a bit of logic involved in it. [Here's the API reference](https://formatjs.io/guides/message-syntax/#select-format).
It would be great if @agrabeli can give feedback about how much of a problem this could be for translators.

hellais: I think this may be a bit overwhelming for translators. Moreover it's a format.js specific, which means it will only work as a translation in JS based project and will not be easy to integrate into things like our apps.

I would suggest, even if that means a bit more LOC, to split this out into individual messages for each state.

Example:

```
{status == "successful" && <FormattedMessage id="Measurement.Details.Website.TCP.ConnectionTo.Successful" />}
{status == "failed" && <FormattedMessage id="Measurement.Details.Website.TCP.ConnectionTo.Failed" />}
{status == "blocked" && <FormattedMessage id="Measurement.Details.Website.TCP.ConnectionTo.Blocked" />}
```

components/footer.js
```diff
@@ -72,32 +73,32 @@ const Footer = () => (
...
-          <FooterHead>Data</FooterHead>
-          <FooterLink href='https://ooni.io/install/' label='Install Probe' />
-          <FooterLink href='https://explorer.ooni.io/' label='Explorer' />
-          <FooterLink href='https://api.ooni.io/' label='API' />
-          <FooterLink href='https://ooni.io/nettest/' label='Test Docs' />
+          <FooterHead>{intl.formatMessage({ id: 'Footer.Headings.Data' })}</FooterHead>
+          <FooterLink href='https://ooni.io/install/' label={intl.formatMessage({ id: 'Footer.Links.Probe' })} />
+          <FooterLink href='https://explorer.ooni.io/' label={intl.formatMessage({ id: 'Footer.Links.Explorer' })} />
+          <FooterLink href='https://api.ooni.io/' label={intl.formatMessage({ id: 'Footer.Links.API' })} />
+          <FooterLink href='https://ooni.io/nettest/' label={intl.formatMessage({ id: 'Footer.Links.Tests' })} />
```
hellais: <Bikeshedding>
Maybe the key name should be singular similar to `Text` in `Footer.Text.Slogan`, as in `Footer.Link.Tests` (instead of `Footer.Links.Tests`).
Same for `Headings`.
</Bikesheedding>

components/measurement/SummaryText.js
```diff
@@ -2,28 +2,45 @@ import React from 'react'
...
+  let textToRender = null
+  if (typeof content === 'function') {
+    textToRender = content()
+  } else {
+    textToRender = <Text py={4} fontSize={3}>
+      <FormattedMessage id='Measurement.Details.SummaryTextTemplate'
+        defaultMessage='The {testName} test
+        in the {network} network of {country} on {date}
+        {content}. Check the network measurement data below, as well as other
+        measurements from this network testing this site to explore further.'
```
hellais: Why does this have a defaultMessage and the others don't?

Is it because it is using tokens we need to have defaultMessage with tokens?

If that is the case to avoid confusion on where the translation comes from I suggest we put a default message that includes the token variables, but doesn't have meaningful text in it.

Maybe something like:
```
EDIT COPY IN CSV {testName} {network} {country}
```

components/measurement/nettests/telegram.js
```diff
@@ -67,13 +50,16 @@ const TelegramDetails = ({ measurement, render }) => {
 
   if (!telegramWebOK || !telegramDesktopOK) {
     anomaly = true
-    hint = 'Telegram is not working'
+    hint = <FormattedMessage id='Measurement.Status.Hint.Telegram.Blocked' />
   }
 
   return (
     render({
       status: anomaly ? 'anomaly': 'reachable',
```
hellais: Is this not being translated? Or are these used as keys to affect how to present some views in other pages?

If the later maybe defining some named constants is more appropriate.

components/footer.js
hellais: I see there are several instances of this. I would suggest a mass search and replace for `.Headings.` with `.Heading.` and `.Links.` with `.Link.` in the code and in the csv.

components/measurement/StatusHeader.js
```diff
@@ -19,19 +20,19 @@ const StatusHeader = ({ status, color, icon, label, info }) => {
   if (status) {
     switch (status) {
     case 'anomaly':
-      computedLabel = 'Anomaly'
+      computedLabel = <FormattedMessage id='Measurement.Status.Header.Websites.Anomaly' />
```
hellais: We should pick either `Header` or `Heading` (see other comment on `Headings`).

I am OK with either.

hellais: I guess `Heading` is most consistent with our design system: https://github.com/ooni/design-system/blob/master/components/atoms/Heading.js

components/footer.js
sarathms: Updated these.

components/measurement/StatusHeader.js
sarathms: The `Header` here refers to the status header region as opposed to headings as text elements. I have used `Headings` for such keys.

hellais: Ah ok I see. Maybe we can call it `Hero` or `Jumbotron` (to take the boostrap terminology) to make it clear that it's a different thing to Headings (we use the `Hero` term on the mobile app).

components/measurement/nettests/telegram.js
sarathms: Assuming you meant _Is this not being used?_:
I wanted to return to this after I restructure the SummaryText component. I will update it accordingly now.

components/measurement/StatusHeader.js
sarathms: I guess I will change it in the file names and other references as well. It is better to follow consistent naming. Sorry about that.

sarathms: These are done.

hellais:
🐳

Issue #20, opened 2018-11-20
Terminate search API query if user navigates to another filter


darkk:
`/search` page does not terminate API query when user switches virtual "tabs". E.g. the following screenshot was reproduced using following sequence of actions:
- open https://ooni-explorer-next.test.ooni.io/search
- wait till "all results" load
- go to "confirmed"
- wait for ~ half a second (API is currently extremely slow for that query)
- switch back to "all results", switch back and forth between "virtual tabs"
- _every_ switch spawns new API query to the backend (e.g. there are several parallel-running `anomaly=true` queries!)
- eventually error is delivered somewhere to the UI, the error may be irrelevant to the "currently open" tab as one can see below (error for anomalous query in "all results" tab):

![no-query-termination 2018-11-16 20-06-40](https://user-images.githubusercontent.com/21046/48793910-7cdac200-ed09-11e8-8c8d-1205ec2ef782.png)

Moreover, as soon as UI enters this state, "Filter results" button (with empty filter) sends valid query to the API, "loading" spinner is shown, API responds with valid response, "loading" spinner is hidden, but the response is not actually rendered (visual state looks exactly the screenshot above: error message, nothing else). Also, the error message is not hidden when the new query is issued while all pending queries are already "finished" (either correctly or failing).

"Bad UX" would just block controls till the request is received and @sarathms suggests that as a last resort. Terminating the running query is, maybe, slightly better pattern.

hellais:
@sarathms can you confirm if this is in fact still the case or not?

sarathms:
Since we changed the UI for the search page, this problem isn't that common anymore. When a query is in progress, the user can work the filters, but those actions don't trigger API requests until the user hits the button.

But when the button is clicked, a new API request is created while the previous one is still active. I think we should terminate the older request at this point and then start the new one.

hellais:
This has been superseeded by https://github.com/ooni/explorer/issues/74

PR #51, opened 2019-02-19
Various fixes to Search Page

### LINKED ISSUE #12: Implement the search page
Here are some items that still need to be done:
* [ ] Add hints and validation to the ASN input box
* [ ] If another search query is done we should terminate previous queries (see: #20)
* [x] Only list countries where we actually have measurements in 
* [x] Populate the list of test_names from a more canonical source (maybe the API?)

[Link to mockup](https://design.ooni.io/mockups/explorer/screen/e84dd189-cece-4495-9982-637ca77eed2d/Search-1/)

Currently the search page is done in part, however it's not fully aligned to the mockups.

In particular some of the form elements are not the same as the mockups. These components should eventually make their way into the `design-system`.

sarathms:
Fixes #12 

## Proposed Changes
(from #12)
* [x] Start date & end date filters don't work as expected (the filtering does not work and the address bar is not rewritten)
* [x] Date filter selectors should validate that the date range is consistent (i.e. start_date < end_date < NOW())
* [ ] Add hints and validation to the ASN input box
* [ ] If another search query is done we should terminate previous queries (see: https://github.com/ooni/explorer/issues/20)
* [ ] Only list countries where we actually have measurements in (@hellais will document how to do this)
* [ ] Populate the list of test_names from a more canonical source (maybe the API?) (@hellais will document how to do this)

### DIFF
--- a/components/Layout.js
+++ b/components/Layout.js
@@ -26,6 +26,7 @@ injectGlobal`
     margin: 0;
     padding: 0;
     font-family: "Fira Sans";
+    font-size: 14px;
     height: 100%;
     background-color: ${theme.colors.white};
   }
--- a/components/Layout.js
+++ b/components/Layout.js
@@ -28,7 +28,7 @@ injectGlobal`
     font-family: "Fira Sans";
     font-size: 14px;
     height: 100%;
-    background-color: ${theme.colors.white};
+    background-color: #ffffff;
   }
 
   /*
--- a/components/footer.js
+++ b/components/footer.js
@@ -9,6 +9,7 @@ const StyledFooter = styled.footer`
   background-color: ${props => props.theme.colors.gray1};
   color: ${props => props.theme.colors.blue9};
   font-size: 0.857rem;
+  margin-top: 32px;
 `
 
 const FooterBox = styled(Box)`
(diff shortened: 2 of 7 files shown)


sarathms:
Added changes made during the session with @elioqoshi. Some previewed changes like container width enhancements have to come in through updates to `ooni-components`.

components/Layout.js
```diff
@@ -26,8 +26,9 @@ injectGlobal`
     margin: 0;
     padding: 0;
     font-family: "Fira Sans";
+    font-size: 14px;
     height: 100%;
-    background-color: ${theme.colors.white};
+    background-color: #ffffff;
```
hellais: As mentioned via voice I think it's fine to not use constants for things like #fff (pure white) or #000 (pure black)

components/footer.js
```diff
@@ -8,7 +8,8 @@ import { injectIntl, intlShape } from 'react-intl'
 const StyledFooter = styled.footer`
   background-color: ${props => props.theme.colors.gray1};
   color: ${props => props.theme.colors.blue9};
-  font-size: 0.8rem;
+  font-size: 0.857rem;
```
hellais: What is the origin of this odd font-size?

hellais:
This looks good to me and I will merge. I will document the remaining issues as part of https://github.com/ooni/explorer/issues/12

components/footer.js
sarathms: When I first wrote the footer, the font size looked a bit big for smaller screens and so ended up using the `rem` notation to keep it relative. When @elioqoshi was reviewing it he found the font-size as `11.2px` where he expected `12px`. But he also agreed that using rem makes sense. Hence this value.

Issue #55, opened 2019-03-22
Show progress indicator when navigating between pages


sarathms:
Something on the lines of [this example](https://github.com/zeit/next.js/blob/master/examples/with-loading/pages/_app.js).

sarathms:
Added in df2e590

elioqoshi:
Is this live @sarathms?

sarathms:
Yes it is live on https://explorer-beta.ooni.io
![image](https://user-images.githubusercontent.com/700829/56303251-8c6c0480-6109-11e9-8a1d-c96b591e0bcb.png)

But you might have to suggest a better color for the loading bar.

hellais:
Yeah I think we should go for some other color as blue on blue barely visible. @elioqoshi @holantonela what do you think?

elioqoshi:
> Yeah I think we should go for some other color as blue on blue barely visible.

Yeah, let's go with the anomaly color (Yellow 9 ) for the loading bar.

hellais:
I have deployed this to: https://explorer-beta.ooni.io/.

@elioqoshi check if it looks good and if so close this issue.

elioqoshi:
LGTM! Closing

PR #79, opened 2019-04-18
Yellow progress bar

hellais:
Fixes: https://github.com/ooni/explorer/issues/55


hellais:
> Instead of copying the entire CSS file, wouldn't be it better to just override background and other such properties and rely on bug fixes and updates to the original module which we still depend on anyway

I am not sure how to do that given that it's a css file. It's also small enough that it didn't seem to matter and vendoring the css is what the docs of nprogress suggest doing in order to customize it, see: https://github.com/rstacruz/nprogress#customization.

sarathms:
It worked when I imported a custom `static/nprogress.css` with only the overrides along with importing the `nprogress/nprogress.css`. But maybe this is indeed an over-engineered, the benefits are hardly measurable. But we should move it to `pages/_document.js` because the nav-bar is designed to be optional to rendering page.

hellais:
lgtm

PR #81, opened 2019-04-18
Update country-util to the latest version

hellais:
This fixes issues with the country name of North Macedonia & Eswatini


hellais:
I pushed again to this branch as the CI is failing due to the auto deploy code not working as expected.

PR #85, opened 2019-04-18
Handle legacy explorer urls

sarathms:
This fixes problems with handling the URLs schema used by the old explorer e.g `https://explorer.ooni.torproject.org/measurement/report-id?input=website-url-tested`

The query string was being omitted when handing off the request to the new URL format. It is not captured and sent to the request handler.


hellais:
Can we do as part of this branch also changes to the search results page to ensure links have consistent measurement page links?

Issue #84, opened 2019-04-18
Tooltips for charts are not usable


hellais:
Currently when I hover over a bar chart in the country page the tooltip is going to be cropped most of the time making the information not visible.

![Screenshot 2019-04-18 at 17 04 27](https://user-images.githubusercontent.com/424620/56370723-0aa8d380-61fc-11e9-9cfb-124d668a7023.png)

We should find a simple way to make at least all the content in them be readable.

elioqoshi:
Could the tool tips also be styled like in the mockups? 

i.e: 
![image](https://user-images.githubusercontent.com/5436686/56371375-4c864980-61fd-11e9-983b-bd6cd48a2918.png)

hellais:
@elioqoshi the tooltip should scale to support at least all the possible values, ie: Total, failures, confirmed, anomalies.

I think that a tooltip is not the right pattern here and I think we should probably have a fixed legend that is updated when you hover over the item.

@sarathms In looking at this bit of code I also noticed there is actually a bug in the rendering of the tooltip. You are registering a `VictoryVoronoiContainer` inside of the `VictoryChart` component. This means that the voronoi container will be triggered on all the stacked bar charts when the chart is all zeros (they are all equally distant from the mouse).

The end result is that the content of the tooltip ends up being duplicated and repeated 4 times.

I did a little bit of research into this matter and the recommended way to deal with tooltips and stacked bar charts is to register the label only on the highest item in the stack, see: https://github.com/FormidableLabs/victory/issues/997#issuecomment-399338369

sarathms:
> bug in the rendering of the tooltip.

@hellais Additionally, that seems to be explain why the tooltip doesn't get triggered in lower parts of the graph. I clearly didn't understand what 'voronoi` means :-|

elioqoshi:
> @elioqoshi the tooltip should scale to support at least all the possible values, ie: Total, failures, confirmed, anomalies.

I should have specified that I meant the style mostly. middle alignment for these kind of values and the container shape (which would scale depending on the content).

> I think that a tooltip is not the right pattern here and I think we should probably have a fixed legend that is updated when you hover over the item.

I do think we should have addressed these in the mockup stage. What is the issue you see with the tooltips and the information needed to be conveyed?

hellais:
> I do think we should have addressed these in the mockup stage. 

Yeah I am sorry for not having addressed this there. It wasn't clear that you were expecting also the on hover effects to be implemented as per the mocks.

[...]

PR #86, opened 2019-04-19
Get the URL charts to look decent

hellais:
Implements tooltips as per the discussion in #84.

I also cleaned up a bit the chart in the overview section, but I could not figure out a way to make the two charts be linked (i.e. so that when you scrub over any of the two graphs the tooltip appears on both of them).


hellais:
@sarathms good point. I opened an issue for that here: https://github.com/ooni/explorer/issues/90

PR #88, opened 2019-04-24
Measurement page ux tweaks

sarathms:
This PR has tweaks to the measurement pages based on comments from @elioqoshi and @holantonela and also from a bunch of issues created earlier.


sarathms:
Reviewed by @hellais. Added minor tweaks based on Antonela's suggestions.

PR #91, opened 2019-04-26
Parse markdown in translated strings

sarathms:
- [x] Add component similar to `<FormattedMessage>` to fetch the translated string and render the markdown content in it

Use the above component to parse markdown in these places:
- [x] Home page intro text
- [x] ~Home Page - highlights~ (will be implemented as part of the highlights page PR)
- [x] Country page - overview
- [x] Country page - Websites description
- [x] Country page - Apps description
- [x] Measurement - Details - SummaryText

### DIFF
--- a/components/formatted-markdown.js
+++ b/components/formatted-markdown.js
@@ -1,0 +1,19 @@
+//
+import React from 'react'
+import PropTypes from 'prop-types'
+import { injectIntl, intlShape } from 'react-intl'
+import Markdown from 'markdown-to-jsx'
+
+const FormattedMarkdown = ({ intl, id }) => (
+  // <Markdown source={intl.formatMessage({ id })} />
+  <Markdown>
+    {intl.formatMessage({ id })}
+  </Markdown>
+)
+
+FormattedMarkdown.propTypes = {
+  intl: intlShape,
+  id: PropTypes.string.isRequired
+}
+
+export default injectIntl(FormattedMarkdown)
--- a/components/formatted-markdown.js
+++ b/components/formatted-markdown.js
@@ -1,4 +1,5 @@
-//
+// Documentation for markdown-to-jsx
+// https://github.com/probablyup/markdown-to-jsx
 import React from 'react'
 import PropTypes from 'prop-types'
 import { injectIntl, intlShape } from 'react-intl'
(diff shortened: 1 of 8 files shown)


components/formatted-markdown.js
```diff
@@ -0,0 +1,20 @@
+// Documentation for markdown-to-jsx
+// https://github.com/probablyup/markdown-to-jsx
+import React from 'react'
+import PropTypes from 'prop-types'
+import { injectIntl, intlShape } from 'react-intl'
+import Markdown from 'markdown-to-jsx'
+
+const FormattedMarkdown = ({ intl, id }) => (
+  // <Markdown source={intl.formatMessage({ id })} />
+  <Markdown>
```
hellais: We may want to make use of https://www.npmjs.com/package/markdown-to-jsx#optionsoverrides---override-any-html-tags-representation so that the styling from our components (for example `<Heading>` and `<Link>`) are used in rendered markdown as well.

On design.ooni.io I was using https://github.com/mdx-js/mdx with a babel loader, but we probably don't want to do that here as supporting jsx in markdown here is not wise as translators could inject malicious JS code and we don't have the need for it.

sarathms: I looked into @mdx-js/mdx, but it seemed more useful for 'JSX in Markdown' which isn't our usecase. So it was between `react-markdown` and `markdown-in-jsx`. This one claimed to have a smaller bundle size. Took their word for it and used it.

sarathms: Removed the commented line from when I was evaluating [`react-markdown`](https://github.com/rexxars/react-markdown).

sarathms:
All planned changes for this PR have been implemented. Please review for merge.

components/formatted-markdown.js
hellais: Did you look into if the styling is going to look consistent in the markdown rendered content?

hellais:
I think it's OK to merge this, but we should maybe open a ticket related to customising the rendered markdown components in case the styling of them is not consistent with our design-system (see: https://github.com/ooni/explorer/pull/91#discussion_r279439016).

sarathms:
Looks like it keeps doesn't meddle with the styling. From a quick glance at the [source code](https://github.com/probablyup/markdown-to-jsx/blob/master/index.js), it doesn't seem to deal with styles unless we are rendering tables. I don't think we need to open an issue right now. We will keep an eye out for UX bugs related to this.

PR #96, opened 2019-05-01
Fix UX issues on country pages

sarathms:
This PR fixes issues pointed out by @holantonela in the below annotated screenshot.

country


sarathms:
I haven't fixed the font-family issue. It needs to be fixed for all charts across the app. The right way forward is to create a theme object that can be applied to all charts. There is a `victoryTheme` in `ooni-components` but applying that messes up the charts quite a bit.

hellais:
I am merging this to work on using the production API.

Open new PRs for the new fixes.

PR #135, opened 2019-05-03
Fix search page

sarathms:


hellais:
I tested it and it works well when you do `yarn run build` and `yarn run start`.

It does not work in dev mode. We should open a ticket for that.

sarathms:
Opened [this ticket in `design-system`](https://github.com/ooni/design-system/issues/52).

sarathms:
And #145 in explorer

PR #151, opened 2019-05-06
Fix measurement summary layout

sarathms:
This PR changes the mobile layout for the measurement summary in the Hero. It now shows a single column of 3 boxes like below.

image


hellais:
This looks good to me.

@holantonela @elioqoshi can you also take a look?

holantonela:
Yes, perfect. Thanks!

PR #162, opened 2019-05-10
Make logo visually clickable

sarathms:
`next/link` doesn't pass the href prop automatically for children other than `<a>`


hellais:
lgtm

PR #163, opened 2019-05-16
Add floating button to collect user feedback

sarathms:


sarathms:
Shows a button that looks like this.
image

On clicking, shows a dialog in a full page overlay.
image

hellais:
lgtm

PR #169, opened 2019-05-22
Fix navigation stickiness

sarathms:


sarathms:
So far
image

There seems to be no CSS-based solution to detect stickiness of elements. We need this to toggle the flag and country code to the left of the navigation bar.

I found [this module](https://github.com/madou/react-sticky-header) which seems to work that way. Will implement this next.

A demo from the project page. 
example

hellais:
In looking into this while working on probe-desktop I found the react-sticky-header to not work so well (not to mention that it wants you to `require` a `css` file), while https://github.com/captivationsoftware/react-sticky works very well out of the box and has much better docs.

I suggest using that.

sarathms:
Made the country page menu responsive close to [mockup here](https://www.figma.com/file/K1NGidHqUldfVuDNmU4MnO/Explorer-Navigation-Menues?node-id=1319%3A1136)

Implemented this way:
image

First use of React Hooks! :1st_place_medal:

components/country/PageNavMenu.js
```diff
@@ -0,0 +1,51 @@
...
+const ToggleIcon = styled(MdExpandLess)`
+  cursor: pointer;
+  background-color: ${props => props.theme.colors.gray3};
+  border-radius: 50%;
+  transform: ${props => props.isOpen ? 'rotate(0deg)': 'rotate(180deg)'};
+  transition: transform 0.1s linear;
+`
+
+const PageNavMenu = () => {
+  const [isOpen, setOpen] = useState(true)
```
hellais: Ah cool. This is the hook pattern I guess.

components/country/PageNavMenu.js
```diff
@@ -0,0 +1,51 @@
...
+          </PageNavItem>
+          <PageNavItem link='#apps'>
+            <FormattedMessage id='Country.Heading.Apps' />
+          </PageNavItem>
+          <PageNavItem link='#network-properties'>
+            <FormattedMessage id='Country.Heading.NetworkProperties' />
+          </PageNavItem>
+        </Flex>}
+      </Box>
+    </React.Fragment>
```
hellais: 👍 
I should probably tweak what I did in probe-desktop to look closer to this. This is much cleaner.

hellais:
While reviewing this I also spotted https://github.com/ooni/explorer/issues/212, which you can do either as part of merging this or a new PR.

PR #170, opened 2019-05-22
Remove repeated headers from HTTP responses

sarathms:
This was added to demonstrate how longer strings in header values are handled. Was committed accidentally.


sarathms:
Fixes this.
image

hellais:
lgtm

PR #182, opened 2019-06-06
Set bigger font size for subtext under measurement status label

### LINKED ISSUE #119: Display IM test status bigger in banner
When you land on the measurement pages for IM tests (e.g. WhatsApp test), the subtitle (in the banner) that communicates whether the specific app is blocked or not is displayed in extremely small writing. For example, "WhatsApp is accessible":

whatsapp-accessible

It would be great if "WhatsApp is accessible" was bigger (same for Telegram & Facebook Messenger tests), so that the user could view it quickly/easily at first glance.

### LINKED ISSUE #168: Measurement Page has wrong title formatting
The subtitle at the apps measurement results is wrong. Let's use the same font-size we have at the websites measurement pages.

Expected
Screenshot_2019-05-21 OONI Explorer(1)

Actual

sarathms:
Closes #168
Closes #119

image

### DIFF
--- a/components/measurement/Hero.js
+++ b/components/measurement/Hero.js
@@ -55,7 +55,9 @@ const Hero = ({ status, color, icon, label, info }) => {
           </Box>
         </Flex>
         {info && <Flex flexWrap='wrap' justifyContent='center'>
-          {info}
+          <Text fontSize={28}>
+            {info}
+          </Text>
         </Flex>}
       </Container>
     </HeroContainer>


sarathms:
Should also remove the `<Text>` component that overrides `input` text size in the `web_connectivity` details component [here](https://github.com/ooni/explorer/blob/fix/extract-keys/components/measurement/nettests/web_connectivity.js#L68).

hellais:
lgtm

PR #183, opened 2019-06-06
Show flag in measurement page hero next to country name

### LINKED ISSUE #116: Display country flag in measurement pages
Perhaps we can add the country flag next to the country name within the measurement pages?

country-flag

That way, when people share screenshots of these pages, you can easily/visually see which country it pertains to (from the image of the flag).

sarathms:
Closes #116


sarathms:
Preview screenshots.

image


image

hellais:
lgtm

hellais:
@holantonela are we sure we want to have a flag next to the country name here?

PR #184, opened 2019-06-06
Remove 'Load More' button when all available results are displayed

### LINKED ISSUE #129: Remove "Load More" button when maximum amount of measurements displayed
When filtering measurements based on a specific date range (in Guatemala, for example), I only get a limited set of measurements, yet the "Load More" button is still displayed (but since we only have 1 measurement in this case, clicking on the "Load More" button wouldn't work).

filter-dates

We therefore need to remove the "Load More" button if the maximum amount of measurements is already displayed.

sarathms:
Closes #129

### DIFF
--- a/components/measurement/CommonDetails.js
+++ b/components/measurement/CommonDetails.js
@@ -55,13 +55,13 @@ const CommonDetails = ({
   } = measurement
 
   let engine_version = 'none',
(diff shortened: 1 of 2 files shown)


sarathms:
We now check the `nextURL` value in the search API response to determine if there are more results to load.

PR #191, opened 2019-06-10
Use expand/collapse icon from MD for collapsible boxes

### LINKED ISSUE #189: Improve UX of the raw measurement data log
I think that the orientation of the expansion collapse button on the measurement page for raw data looks a bit odd.

Screenshot 2019-06-08 at 13 30 51

Moreover it doesn't give any response when you click on it to expand it.

We should see if we can improve the UI of this.

Also is there a reason to not have the raw data expanded by default instead of requiring the user to expand them? It is at the end of the page so I don't think it should pose a problem to have it expanded by default as it is in the current explorer.

cc @holantonela

sarathms:
Closes #189


sarathms:
Deployed for review.

PR #193, opened 2019-06-11
Show confirmed and failed measurements correctly

### LINKED ISSUE #36: ‘confirmed blocking’ measurements
Currently it's not possible to determine the "confirmed" flag of measurements from the raw measurement.

This flag is only available in the search listing of the API. We should come up with a way of delivering annotations from the pipeline stored in the metadb to the explorer measurement pages.

### LINKED ISSUE #131: Confirmed blocked pages should have red banners, instead of orange
The Confirmed Blocked (Web Connectivity) measurement pages should have **red** banners, instead of orange ones (which are used for anomalous measurements).

Examples: http://explorer-beta.ooni.io/measurement/20190501T195957Z_AS197207_UEwHpRQ5zbd2lNSwcr5XfKxxJTsizM67Z4mwqEyxtoHcecR9ft?input=http://www.roozonline.com/ & http://explorer-beta.ooni.io/measurement/20190501T085412Z_AS42668_zDAYOOjqswLhfCTZE12M8TYKXDY0BgP8RWTi10wY6R6OXwC9To?input=http://kinokubik.org/

### LINKED ISSUE #154: Web Connectivity test status undetermined
## Expected Behavior
Measurement page for `web_connectivity` should determine if the 'status' is `accessible`, `anomaly`, `confirmed` or `failure`. And display messages to convey the necessary details in the Hero and Summary Text areas.

## Actual Behavior
Unable to determine the status. Summary text is blank as the conditions checked are not met. The fact that all measurements currently flag `isFailure = true` in metadata, doesn't help in checking if this is indeed a failed test.

  1. [This measurement](http://explorer-beta.ooni.io/measurement/20190504T152614Z_AS6830_n89RPFJzbMeQGsnxoKTMtoPg5xzUlRvguxwh7IT1F7vfcnksxE?input=http://www.radioislam.org/)

### LINKED ISSUE #178: Show website is down on Web Connectivity measurement page
Here is an example of it:
http://explorer-beta.ooni.io/measurement/20190602T034152Z_AS29032_sFjEu7kwoDsBul2dsW6Pqd5ykiAnP4fEQwoGTZVsUmAm7rwiHG?input=http://www.misseastafrica.co.uk

The value of the blocking key is `false`, so it should not be an anomaly:

```
"blocking": false
```

This may be due to the site being marked as `accessible: false`.

sarathms:
Closes #36 
Closes #131 
Closes #154
Closes #178


sarathms:
Measurements to test:
(from #131)
http://explorer-beta.ooni.io/measurement/20190501T195957Z_AS197207_UEwHpRQ5zbd2lNSwcr5XfKxxJTsizM67Z4mwqEyxtoHcecR9ft?input=http://www.roozonline.com/
http://explorer-beta.ooni.io/measurement/20190501T085412Z_AS42668_zDAYOOjqswLhfCTZE12M8TYKXDY0BgP8RWTi10wY6R6OXwC9To?input=http://kinokubik.org/

(from #154)
http://explorer-beta.ooni.io/measurement/20190504T152614Z_AS6830_n89RPFJzbMeQGsnxoKTMtoPg5xzUlRvguxwh7IT1F7vfcnksxE?input=http://www.radioislam.org/

hellais:
I think this looks good, but I am dubious about two things:

1. I think the color contrast in the case of the yellow is too low. It is 1.6 which is less than the recommended 2 minimum.
Screenshot 2019-06-13 at 18 07 54
I suggest we either invert the color of the text (use black instead of white) or we use a different color, but we should use at minimum #f59f00 (yellow7) which is closer to orange than yellow actually.

2. I think we should use a different icon from the confirmed icon for failure. Perhaps we can use the question mark `?` as since the measurement failed we cannot say anything

3. I think we need to drop the flags from the country because certain flags look bad with some of the colors

cc @holantonela

sarathms:
You are right. The mockup actually had the text in black for failed measurements.
image

It is not straight-forward to re-style the svg logo for select contexts. I will research it if we decide that we cannot use a different color. Rest can be done as shown below.
image

hellais:
If it's not simple to restyle it inside of react maybe as a workaround we can just make an alternative version of the icon which is white?

Though I suggest we use a different icon than the `X` to avoid confusing it with the confirmed icon.

holantonela:
We agreed on using `grey6` (#868e96) as background for error pages. The [?] icon is good to go, too. After it, this PR can be merged.
Screenshot_2019-06-17 OONI Explorer

hellais:
This doesn't seem to be implemented as the mocks yet.

sarathms:
Links to verify completion of this PR
[Accessible](https://explorer-beta.ooni.io/measurement/20190721T115930Z_AS22394_B7TvvPCYku77xGxNnUrOHii30PAdZ2rQNeCnRzI0h1FjGLJ20k?input=http://www.pdhre.org/)
[Anomaly](https://explorer-beta.ooni.io/measurement/20190721T115930Z_AS22394_B7TvvPCYku77xGxNnUrOHii30PAdZ2rQNeCnRzI0h1FjGLJ20k?input=http://whitehonor.com/)
[Confirmed](https://explorer-beta.ooni.io/measurement/20190704T001123Z_AS39891_fA5Z1yL899xCMUD7P8hWasDbyr2p8JSzFlzznX2ZpkZhijm5bU?input=http://www.mossad.gov.il/default.aspx)
[Error](https://explorer-beta.ooni.io/measurement/20190504T152614Z_AS6830_n89RPFJzbMeQGsnxoKTMtoPg5xzUlRvguxwh7IT1F7vfcnksxE?input=http://www.radioislam.org/)
[Site Down](https://explorer-beta.ooni.io/measurement/20190602T034152Z_AS29032_sFjEu7kwoDsBul2dsW6Pqd5ykiAnP4fEQwoGTZVsUmAm7rwiHG?input=http://www.misseastafrica.co.uk)

What is pending:
- [ ] A custom **`?`** icon for error pages. Currently uses the one in `react-icons/lib/md`
- [ ] A suitable icon for `Website Down` pages. Uses the same as error page for now.
- Copy for website down pages (which will be done along with other missing ones separately)

hellais:
@sarathms 
>  A custom ? icon for error pages. Currently uses the one in react-icons/lib/md

I suggest you use the `FaQuestion` icon which is already part of react-icons

>  A suitable icon for Website Down pages. Uses the same as error page for now.

I suggest you use `MdWarning` for this page and that you change the color of this page to be the same as the error page (gray).

sarathms:
> I suggest you use the `FaQuestion` icon which is already part of react-icons

:+1: 

> I suggest you use `MdWarning` for this page and that you change the color of this page to be the same as the error page (gray).

We discussed in [this thread](https://github.com/ooni/explorer/issues/178#issuecomment-498305333) to go for greeen. But I agree that it is better to use gray.

hellais:
LGTM.

PR #196, opened 2019-06-14
Extract react-intl keys

### LINKED ISSUE #180: Sync the master spreadsheet for strings to the latest
This should be updated: https://docs.google.com/spreadsheets/d/1QAeDVF5wYj7bnTE6rJ0XX3C8sBF0o6A0TxeLw9fpTvM/edit#gid=0

sarathms:
Adds a script to extract all keys from the codebase into a file that can be used by the [`json-key-check.py` script](https://github.com/ooni/translations/blob/explorer/json-key-check.py). This uses `babel-plugin-react-intl` which doesn't work for `<FormattedMessage>` components which use dynamic ids e.g 
```
<FormattedMessage id=`Measurment.Label.BlockingReason.${reason}`  />
```
because babel [fails to statically evaluate such ids](https://github.com/formatjs/babel-plugin-react-intl/issues/119).

To work around this, we have to refactor such declarations using a combination of `defineMessages()` and `intl.formatMessage()` which in turn forces us to define all the possible messages upfront for the babel plugin to extract statically.

Closes #180


sarathms:
I am yet to refactor `components/search/results-list.js` to follow the above pattern. It uses a `Search.${testName}.Resutls.${status}` structure which results in more than 25 keys that need to be explicitly defined. :man_shrugging:

hellais:
Yeah the dynamic resolution of ids seems a bit annoying.

It seems like this branch now needs some conflict resolution. I will finish the review once it's done, but at a glance it seems reasonable to me.

sarathms:
> Yeah the dynamic resolution of ids seems a bit annoying.

I tried to avoid this route for quite some time. Nothing else seems to work if we want to use codebase as the source of truth. As of this PR, the workflow is:
1. `yarn build`
2. `node scripts/extract-react-intl-keys.js` -> ./static/lang/keys.json
3. Download spreadsheet as csv
4. `~/ooni/translations $  ./explorer-sync-csv-source.sh` -> explorer/en/strings.json
5. `python json-key-check.py --json-strings explorer/en/strings.json --json-code ../explorer/static/lang/keys.json`
6. Finally, to build the translations: `cp ../translations/explorer/en/strings.json static/lang/en.json && yarn run build-translations`

> It seems like this branch now needs some conflict resolution.

Yeah that was expected. Will resolve now.

sarathms:
@hellais This is mostly done. The keys to render `ResultTag`s for `http_requests` tests were missing in the copy. Do we set `msmt.anomaly`, `msmt.confirmed` for these tests as well? If we do, then I will define messages for them as well.

hellais:
@sarathms yes we also confirm measurements from `http_requests` tests

PR #199, opened 2019-06-18
Contextual search filters

### LINKED ISSUE #52: Make search filters context-aware
## Expected Behavior
Fields like `Input` in the search page filter sidebar should not be active/effective when used for certain tests like WhatsApp, DASH etc.

## Actual Behavior
When a search term is included in `Input` field, it fails to show the measurements which match the rest of the search criteria. 

  1. Open search page
  2. Select `Telegram` under `Test Name`
  3. _Optionally, select a country with fewer measurements for faster response._
  4. Hit 'FIlter Results' to see some measurements listed.
  5. Enter any text in the `Input` field
  6. Empty results.

### LINKED ISSUE #83: Disable the input form box when a test that doesn't support an input is selected
I think that when the user chooses a test which doesn't support an input (example: WhatsApp) we should be removing (or disabling) the input form as it doesn't make sense to filter by input there.

### LINKED ISSUE #138: Filter results should be greyed out when no changes are made to the filter
Screenshot 2019-05-03 at 19 16 45

I think it would be more usable if we made the filter result button change state to being clickable only if changes are made to the selections.

sarathms:
Closes #52 
Closes #83 
Closes #138

### DIFF
--- a/components/search/filter-sidebar.js
+++ b/components/search/filter-sidebar.js
@@ -50,7 +50,7 @@ class FilterSidebar extends React.Component {
     this.state = {
       inputFilter: props.inputFilter || '',
       onlyFilter: props.onlyFilter || 'all',
-      testNameFilter: props.testNameFilter || '',
+      testNameFilter: props.testNameFilter || 'XX',
       countryFilter: props.countryFilter || '',
       asnFilter: props.asnFilter || '',
       sinceFilter: props.sinceFilter || '',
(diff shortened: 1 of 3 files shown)


sarathms:
Although this is functional, the button component from `ooni-components` doesn't have styling to represent the `disabled` state.

It looks for a dirty state of the search form, but doesn't detect a reset/undo action. e.g when changing the test name from 'NDT" to 'Dash' and back to 'NDT', the form is considered to be dirty.

sarathms:
Deployed.

hellais:
It seems to be working for me only on first load.

For example if I do:

1. Open the search page (as expected the Filter button is disabled)
2. Choose Italy from the country list (as expected the filter button is now clickable)
3. Click on filter
4. When the page loads the filter button is still clickable though I haven't made any changes to it

sarathms:
I mistook the button handler for `setState` earlier. Setting dirty state to false explicitly now.

hellais:
> I mistook the button handler for setState earlier. Setting dirty state to false explicitly now.

Is this deployed? I still see the same behaviour on the branch.

sarathms:
Deployed now.

hellais:
Yes this is good. Merging.

PR #205, opened 2019-06-25
Fix charts

### LINKED ISSUE #90: Fix overlap of tooltips in the country overview
As mentioned by @sarathms in https://github.com/ooni/explorer/pull/86.

> the tooltips for the two charts in the svg now overlap at most places

In the case I used as an example they weren't overlapping as much, but I can definitely see how if the data is quite stationary the tooltips can overlap quite often.

I would say that we deal with it by either:
* Making some tweaks to the style and positioning of the networks tooltip (ex. we don't need to repeat the date in there and we can bottom align it)
* Consolidate all the info just one tooltip

I prefer the first option because keeping the tooltips separate we are teaching the user what the line graph means.

cc @holantonela @elioqoshi

### LINKED ISSUE #101: Add vertical bar or dot on chart in the landing page Monthly coverage worldwide chart
Currently the Monthly coverage worldwide chart doesn't have an indication of which datum on the line chart you have selected when you scrub over the chart.

Screenshot 2019-05-02 at 12 31 20

We should add either a vertical bar aligned to the date the user has selected or maybe make a dot appear on the line in the position currently selected by the user.

cc @holantonela @elioqoshi

### LINKED ISSUE #104: Tooltips on the websites charts are too small
Screenshot 2019-05-02 at 12 53 54

### LINKED ISSUE #150: Country page - Link graphs to measurements
The graphs (under the Websites and Apps sections) on the country pages are really cool, but they're not very actionable at the moment. 

For example, when you land on the Iranian page, you see that a number of websites presented blocking in the timeline of their testing (the red bars displayed in the graph)...

ir-blocked-sites

...but when you click on the red bars, nothing happens, and you can't easily access the measurements to learn more.

I realize that the graphs present aggregates, but it would be great if the chart was more interactive. 

Specifically, it would be great if, by clicking on bars on the chart, the user is re-directed to the relevant measurements on the Search page. And it would be great if the same applied to the graphs under the Apps section as well.

### LINKED ISSUE #165: Long URLs in websites section on Country page
## Expected Behavior
Long URLs should be truncated or wrapped around to the next line, especially on smaller screens.

## Actual Behavior
Overflows to the right of the page and messes up the page layout.

### LINKED ISSUE #167: Consider hiding charts when on mobile if they are unusable
We should probably hide some of the bigger charts (like the first one on the country page) when the user is via mobile as they will not be usable and they take up precious screen real-estate.

The `responsive` tag should also be removed from charts as it leads to small tooltips (see: https://github.com/ooni/explorer/issues/104)

Related: https://github.com/ooni/explorer/issues/101 https://github.com/ooni/explorer/issues/150

### LINKED ISSUE #174: Country pages: apps - UI adjustments
- [x] Review the padding at the table header. "Last tested + icon" should be aligned to the right.
 Apps Table Header
- [x] The padding between each row should be reduced by 50%.
- [x] For messaging apps with less than 10 measurements should have the toggle open by default.
- [x] Each ASN per row should link to a search result for that ASN + that test query, like `/search?probe_asn=AS29032&test_name=facebook_messenger`. This will fix https://github.com/ooni/explorer/issues/150
- [ ] The bars are showing qty, we could change the color to black/darkgrey rather than green.

sarathms:
Includes a custom theme used on all the graphs across the site. Hence contains fixes to multiple issues related to charts.

Closes #101 
Closes #90 
Closes #104 
Closes #167
Closes #174
Closes #150 
Closes #165


hellais:
Does this branch fix also the tooltip font size being too small?

sarathms:
It is expected to, but haven't achieved yet. 

Clearly, I used a weird way to annotate that in the first comment. Updated it now.

hellais:
This looks good. I am going to merge this PR and suggest we implement the remaining things for:
* https://github.com/ooni/explorer/issues/101
* https://github.com/ooni/explorer/issues/104
* https://github.com/ooni/explorer/issues/167
* https://github.com/ooni/explorer/issues/174

In a new PR.

PR #209, opened 2019-07-04
Cards-based highlights on landing page

### LINKED ISSUE #200: Improve highlights
Besides searching, highlights listing is best feature Explorer has. We can improve the way we are displaying this section on our homepage.

Using [the data set](https://github.com/ooni/explorer/blob/master/components/landing/highlights.json), I worked in two different layouts for this section. Cards are an excellent way to show this information. We can have a carousel with tabs to navigate different sets (1), or we can have different carousels for each listing set (2).

The card background color is taking [the darkest hex](https://stackoverflow.com/questions/12043187/how-to-check-if-hex-color-is-too-black) available as a `fill` in the flag.svg.

What do you think?

1.

2.

sarathms:
Closes #200


sarathms:
It is not responsive right now. Changing the number of visible cards to `1` should make a significant difference.

sarathms:
Used the `useEffect` hook to detect mobile layout. We could probably extract and use it as a general hook [like this one here](https://github.com/haldarmahesh/use-mobile-detect-hook)

hellais:
The cards look pretty different in the implementation than in the mock.

I am personally not a big fan of carousels, because I find that they hide the information that is available and are not very usable (especially on mobile).
It also seems like for two of the rows we only have 1-2 extra items to scroll on the side to making the carousel not so 

@holantonela can we perhaps find a way to adapt the style you suggest in #200 work without a carousel and just be static?

holantonela:
yes, we agreed on use the grid and remove the carousel. More info at #200.

sarathms:
As a grid now:
image


Mobile
photo_2019-07-19_15-26-46
photo_2019-07-19_15-26-50

hellais:
@sarathms can you deploy this onto explorer-beta?

hellais:
I ran this in development, and I think they are way too big now. I will try to make some tweaks to the sizes in this branch and report back.

hellais:
I made some tweaks to the font sizes to make it flow better with the rest of the page here: https://github.com/ooni/explorer/pull/209/commits/e14fd19deb7fbad3c8fc7faba891afdbf87581e1.

**BEFORE**
localhost_3100_ (1)

**AFTER**
localhost_3100_


I also changed the style of section heading to be the same style as that of sections in the beginning of the landing page ("Blocking of Websites & Apps", "Search", etc.).

@holantonela what do you think?

sarathms:
This is now [deployed](https://explorer-beta.ooni.io/).

hellais:
I am going to merge this, but input from @holantonela on this is still useful.

PR #210, opened 2019-07-04
Vanilla Tor stats on country page

### LINKED ISSUE #105: The vanilla_tor section should show a zero state (or not displayed) when data is missing
For example if I go to https://explorer-beta.ooni.io/country/BF, which has no vanilla_tor measurements I still see that section saying that there is a measurement from "a few seconds ago", yet the row is not clickable.

Screenshot 2019-05-02 at 12 56 16

sarathms:
Closes #105


hellais:
lgtm

Issue #204, opened 2019-06-24
Multiple values on y-axis for smaller ranges


sarathms:
When number of measured networks is within a small range of say `[0, 4]`, the y-axis tick values end up as duplicates since we round them off to the nearest whole number.
![image](https://user-images.githubusercontent.com/700829/60038002-19d75200-9681-11e9-972b-78183d7e7a22.png)

This can be eliminated by rounding to one decimal point, but it wouldn't make much sense on the graph. Since we are talking about number of measurements.

hellais:
I am going to look into fixing this.

hellais:
I opened a PR with a fix for this in: https://github.com/ooni/explorer/pull/216.

PR #216, opened 2019-07-18
Manually compute the tickValues based on the maxima.

hellais:
By doing it manually we are able to ensure that the values we tick to
are always whole numbers and are always 4.

This fixes: https://github.com/ooni/explorer/issues/204

### DIFF
--- a/components/country/overview-charts.js
+++ b/components/country/overview-charts.js
@@ -79,9 +79,10 @@ class TestsByGroup extends React.PureComponent {
     const testCoverageByDay = testCoverage.reduce((prev, cur) => {
       prev[cur.test_day] = prev[cur.test_day] || {}
       prev[cur.test_day][cur.test_group] = cur.count
+      const allTestCount = Object.values(prev[cur.test_day]).reduce((p,c) => p+c, 0)
       if (typeof testCoverageMaxima === 'undefined'
-          || testCoverageMaxima < cur.count) {
-        testCoverageMaxima = cur.count
+          || testCoverageMaxima < allTestCount) {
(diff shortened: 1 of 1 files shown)


components/country/overview-charts.js
```diff
@@ -79,9 +79,10 @@ class TestsByGroup extends React.PureComponent {
     const testCoverageByDay = testCoverage.reduce((prev, cur) => {
       prev[cur.test_day] = prev[cur.test_day] || {}
       prev[cur.test_day][cur.test_group] = cur.count
+      const allTestCount = Object.values(prev[cur.test_day]).reduce((p,c) => p+c, 0)
       if (typeof testCoverageMaxima === 'undefined'
-          || testCoverageMaxima < cur.count) {
-        testCoverageMaxima = cur.count
+          || testCoverageMaxima < allTestCount) {
+        testCoverageMaxima = allTestCount
```
hellais: While I was at it I also fixed this issue in calculating the maxima of the testCoverage (we were only considering the maxima of the element which had the most number of measurements and not all the elements).

sarathms: :+1:

hellais:
> When the grid on the chart is enabled (as in styled), it seems to render them in different alignments which seems to make sense considering the tick values are rounded to different levels

Yes you are right. I also noticed it while developing it, but concluded that there is really no way around that. We either draw the ticks in such a way where they align perfectly, but on fractional values or we draw them on whole numbers but they will not align perfectly between the left and right axis.

I personally prefer to have them be precise. As you mention in fix-charts we drop these so it should not be a problem.

PR #230, opened 2019-07-30
Fix layout of common details on measurement pages

sarathms:
Fix the layout mess in the common details at the bottom of measurement pages.
image


hellais:
lgtm

PR #231, opened 2019-07-30
Check for valid tcp_connect key in whatsapp msmt

### LINKED ISSUE #229: Handle missing tcp_connect data in WhatsApp measurements
Measurement in question [is this](https://explorer-beta.ooni.io/measurement/20190620T081634Z_AS24757_6JefVaUG2C4Drg8jQlzfOAXC2PB3KGChehEcfdne3ZHYsM7YbP)

@hellais Shouldn't it have an false-y value like `null` or `false` instead of dropping the key entirely? Or is this a one-off case?

sarathms:
Closes #229


hellais:
Yes this is good.

PR #234, opened 2019-08-01
Show and use current date in default search page

### LINKED ISSUE #232: Default to choosing the current date in the search
Currently when you go to the search page the until filter is left blank and hence even measurements from the future appear in the results.

I think it would be ideal to pre-fill the until field with the current date. This has the added benefit that if somebody shares a search listing result with somebody it will most likely be valid in the future as well (as opposed to the list changing depending on the current date).

sarathms:
Fixes #232

### DIFF
--- a/components/search/filter-sidebar.js
+++ b/components/search/filter-sidebar.js
@@ -11,6 +11,7 @@ import {
   RadioGroup,
   RadioButton
 } from 'ooni-components'
+import moment from 'moment'
 
 import DatePicker from '../date-picker'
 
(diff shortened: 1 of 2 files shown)


hellais:
Can you also populate the URL bar with the current date inside of the until field?

sarathms:
> Can you also populate the URL bar with the current date inside of the until field?

Added commits for this. Deployed to beta.

hellais:
Yes this looks good. Merging.

PR #237, opened 2019-08-01
Make ASN Selector a controlled component

### LINKED ISSUE #236: ASN switcher on country page is broken
## Expected Behavior
When an ASN is selected from the dropdown, it should be the selected one after the page is updated with new data.

## Actual Behavior
It always shows the first item in the list as the selected one, although data shown is for the selected ASN.

sarathms:
Closes #236


hellais:
Is this currently deployed? I still see the bug in the deployed beta.

hellais:
Ok yes now it works as expected! Merging!

PR #238, opened 2019-08-01
Show raw measurement box open by default

### LINKED ISSUE #235: The raw measurement data section should be open by default
Currently the Raw Measurement Data section is collapse when you open the page. It should be expanded by default. I think this is the right pattern as it is at the end of the page and it does not clutter the UI having it expanded by default.

sarathms:
Fixes #235


hellais:
lgtm

PR #239, opened 2019-08-02
Show anomaly or blocking reason in hero

### LINKED ISSUE #179: Re-add Measurement.Hero.Status.Anomaly into Hero unit for Measurement
Currently the Hero Unit is missing the strings from `Measurement.Hero.Status.Anomaly.*` next to the Anomaly string.

sarathms:
Closes #179 
Now we show reason in Hero like so:
image

Works on confirmed blocked pages too:
image
 
I found an exception in this measurement. It is flagged with `isConfirmed = true` because it does serve a block page. But it shows up [normal on old explorer](https://explorer.ooni.io/measurement/20190703T163625Z_AS8708_UwLHVRiAoYGcScetNQ2uobFjyawbW11oqw4ZMIN8qJEZEF41pX?input=http:%2F%2Fwww.slotland.com). Because `test_keys.blocking = false`, it doesn't pick up from the map of reasons we use otherwise. Is there a way to determine reasons for such measurements? Just apply `blockingReason.http-diff` to all confirmed ones?

https://explorer-beta.ooni.io/measurement/20190703T163625Z_AS8708_UwLHVRiAoYGcScetNQ2uobFjyawbW11oqw4ZMIN8qJEZEF41pX?input=http://www.slotland.com

### DIFF
--- a/components/measurement/nettests/web_connectivity.js
+++ b/components/measurement/nettests/web_connectivity.js
@@ -367,7 +367,7 @@ const WebConnectivityDetails = ({
     )
   } else if (isAnomaly) {
     status = 'anomaly'
-    reason = reasons[blocking]
+    reason = intl.formatMessage(messages[`blockingReason.${blocking}`])
     summaryText = (
       <FormattedMessage
         id='Measurement.SummaryText.Websites.Anomaly'
@@ -439,7 +439,7 @@ const WebConnectivityDetails = ({
     <React.Fragment>
       {render({
         status: status,
-        statusInfo: <StatusInfo url={input} />,
+        statusInfo: <StatusInfo url={input} message={reason || null} />,
         summaryText: summaryText,
         details: (
           <React.Fragment>
--- a/components/measurement/nettests/web_connectivity.js
+++ b/components/measurement/nettests/web_connectivity.js
@@ -49,50 +49,6 @@ const messages = defineMessages({
   },
 })
 
-export const checkAnomaly = ( testKeys ) => {
-  const {
-    accessible,
-    blocking,
-  } = testKeys
-
-  let anomaly = null
-  let hint = <FormattedMessage id='Measurement.Status.Hint.Websites.NoCensorship' />
-
-  if ((accessible === true || accessible === null) && blocking === null) {
-    hint = <FormattedMessage id='Measurement.Status.Hint.Websites.Error' />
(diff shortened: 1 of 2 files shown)


hellais:
I suppose this also needs to be deployed to the beta instance.

components/measurement/nettests/web_connectivity.js
```diff
@@ -49,53 +49,9 @@ const messages = defineMessages({
...
-      hint = <FormattedMessage id='Measurement.Status.Hint.Websites.TCPBlock' />
-    }
-  }
-
-  return {
-    status: anomaly,
-    hint
-  }
-}
-
```
hellais: I don't see this function call getting removed from this PR. Was this already dead code that is removed just now?

sarathms: Yes, it was already removed in an earlier PR.

sarathms:
This was already deployed. You probably only checked the link in the first comment which was an exception. Here are a few working examples:
* [anomaly](https://explorer-beta.ooni.io/measurement/20190807T002539Z_AS46562_8uDD0mTLSSLSkjoJvwYWyILuJqOTw53T5vA9Q42u2ElDp8yn2s?input=http://www.greenpeace.org/)
* [confirmed](https://explorer-beta.ooni.io/measurement/20190704T001123Z_AS39891_fA5Z1yL899xCMUD7P8hWasDbyr2p8JSzFlzznX2ZpkZhijm5bU?input=http://www.mossad.gov.il/default.aspx)

hellais:
> This was already deployed. You probably only checked the link in the first comment which was an exception.

Ah yes you are right. The other ones work well.

Why does this one not show any message under it: https://explorer-beta.ooni.io/measurement/20190703T163625Z_AS8708_UwLHVRiAoYGcScetNQ2uobFjyawbW11oqw4ZMIN8qJEZEF41pX?input=http://www.slotland.com?

hellais:
Ok I know why that one does not show any message under it, it's due to the fact that the control request failed and hence the blocking key value is set to false.

I guess it's probably OK for the moment to leave this as is, however I filed an issue so we don't forget about this edge case and consider improving the pattern for it here: https://github.com/ooni/explorer/issues/248

hellais:
For me this is good to merge.

PR #244, opened 2019-08-08
Hide indicators in search results

### LINKED ISSUE #240: Remove the accessible side tab from every result row in search except web_connectivity
Since we don't calculate pipeline side heuristics for the IM tests or any other test than web_connectivity we should drop the green side bar from these result rows in search.

sarathms:
Closes #240

### DIFF
--- a/components/search/results-list.js
+++ b/components/search/results-list.js
@@ -336,38 +336,6 @@ ResultItem.propTypes = {
   msmt: PropTypes.object
 }
 
-const LegendColorBox = styled.div`
-  width: 20px;
-  height: 20px;
-  border-radius: 50%;
-  background-color: ${props => props.color};
-  float: left;
-  margin-right: 10px;
-`
-
-const StyledLegendItem = styled.div`
-  float: left;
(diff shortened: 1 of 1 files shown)


sarathms:
1. This makes the search results 'table' a little bit more empty on the right side. Especially when filtering by non-website tests and even otherwise with dominance of NDT tests. If relevant, we should review this in the usability study.

2. How important is it now for us to have this legend on top of the result list?
image

hellais:
Sharing a screenshot for reference:

Screenshot 2019-08-09 at 11 11 09

> This makes the search results 'table' a little bit more empty on the right side.

I see what you mean with this, but I don't think it's a problem from the perspective of the layout. Whitespace on the right is I think perfectly fine.

> How important is it now for us to have this legend on top of the result list

I think that given that we expand the label inside of the row itself, we should drop the legend from the top especially given that it's only really relevant to the web_connectivity test results currently.

hellais:
This looks good to me. Please merge when you see fit.

PR #245, opened 2019-08-09
Fallback to english for unsupported languages

### LINKED ISSUE #228: When the browser language is unsupported, default to english
Currently if the user has their browser language set to something other than english we will show the to be translated strings. We should instead default to english for an unsupported language.

sarathms:
Closes #228


components/withIntl.js
```diff
@@ -29,11 +29,15 @@ const withIntl = (Page) => {
...
   return class PageWithIntl extends Component {
     render () {
-      const locale = getLocale()
       const now = Date.now()
-
+      let locale = getLocale()
+      // Use 'en' when locale is unsupported
+      if (Object.keys(messages).indexOf(locale) < 0) {
+        locale = 'en'
+      }
```
hellais: I think that currently, since english is the only supported language, we can just hardcode it to english directly. What do you think?

sarathms: I don't have a strong opinion on that. It is more or less the same thing. `messages` is an object with single property, so it doesn't take a lot of cycles to do this step either.

On the other hand, when we do roll out more language support, we will probably want to do the whole language negotiation differently. So this code will be thrown out anyway.

hellais: Ok yeah that seems reasonable. I figured I would just point it out, so we don't by mistake add incomplete locales to wherever it's drawing this from.

PR #251, opened 2019-08-13
Add a custom Error Page

### LINKED ISSUE #56: Show better 404 page for invalid measurement pages
## Expected Behavior
When an invalid `report_id` is passed into the measurement page, we should convey it clearly to the user.

## Actual Behavior
Just shows a simple message on blank page.

sarathms:
Closes #56 

image


pages/_error.js
```diff
@@ -0,0 +1,112 @@
...
+    return (
+      <Layout>
+        <Head>
+          <title> Unknown Error </title>
+        </Head>
+        <NavBar />
+        <Container>
+          <FullHeightFlex alignItems='center' justifyContent='center'>
+            <Heading h={4}>
+              There was an unknown error from our end.
```
hellais: Can we do something in here to emit some more useful debug information with more details about the error?

Even just the `error.toString()` I think would be useful.

pages/_error.js
```diff
@@ -0,0 +1,112 @@
...
+      </Layout>
+    )
+  }
+
+  render () {
+    const { errorCode } = this.props
+    let content = null
+    if (errorCode === 404) {
+      content = this.render404()
+    } else {
```
hellais: Maybe it's more clear if you omit the `else` statement and just assign content to `this.render500()`.

I initially thought there was some un-handled code path, but actually the last return statement is unreachable code.

pages/_error.js
sarathms: Let me check if all such cases trigger a sentry event. In which case, we need not bother users with unnecessary details.

pages/_error.js
sarathms: I didn't catch you. I am not returning in the `if` block. Do you mean
```
let content = this.render500()
if (404) {
  content = this.render404()
}
return content
```

How is the last statement unreachable?

hellais: Oh yes you are right, it is not unreachable.

I guess the problem with what you propose there is that you call the `render500()` in every case which is not ideal.

Maybe it could be refactored like:

```
if (404) {
  return this.render404()
}
return this.render500()
```

sarathms: > you call the render500() in every case 
Which is why I went with the approach I took. Initially, I had a `switch` statement with cases for `404`, `500` and `default`. Then merged `500` and `default` and then I figured a simple if-else should just do the job. The `content =` is a remnant of that. We can get rid of that too.

pages/_error.js
hellais: I think even if we do log things via sentry it's still useful to show messages to users so that they feel empowered to use the error message to report the bug to us (or check on the issue tracker in case the bug has already been noticed by us).

If we don't do that we run the risk of having users send us emails complaining about bugs and us not knowing exactly what is the bug this particular user is experiencing.

pages/_error.js
hellais: Yeah sounds good.

pages/_error.js
```diff
@@ -0,0 +1,109 @@
...
+          <title> Unknown Error </title>
+        </Head>
+        <NavBar />
+        <Container>
+          <FullHeightFlex alignItems='center' justifyContent='center' flexDirection='column'>
+            <Heading h={4}>
+              There was an unexpected error from our end.
+            </Heading>
+            <Text my={4}>
+              Maybe try <Link href='/countried'><a>exploring some measurements</a></Link> or go to our <Link href='/'><a>homepage</a></Link>.
```
hellais: Typo see: https://github.com/ooni/explorer/issues/255

pages/_app.js
```diff
@@ -68,29 +69,7 @@ export default class MyApp extends App {
...
-            href='#'
-            onClick={() => {
-              window.location.reload(true)
-            }}
-          >
-            Or, try reloading the page
-          </a>
-        </p>
-      </section>
+      <ErrorPage errorCode={500} />
```
hellais: Maybe what you can do is pass in here the `errorEventId` and the `errorInfo` to the `ErrorPage`?

pages/search.js
```diff
@@ -61,7 +76,7 @@ const formatError = (error) => {
     errorString += ` (${error.errno})`
   }
   if (errorString === '') {
-    errorString = JSON.stringify(error)
+    errorString = JSON.stringify(error, getCircularReplacer())
```
hellais: Why are we getting circular references here? Maybe we should open an issue on zeit/next about it?

PR #254, opened 2019-08-14
Add zero state for missing data

### LINKED ISSUE #215: Add zero state for when data is missing
A good example of this is: https://explorer-beta.ooni.io/country/IO.

sarathms:
Closes #215

### DIFF
--- a/components/country/overview-charts.js
+++ b/components/country/overview-charts.js
@@ -72,42 +72,154 @@ class TestsByGroup extends React.PureComponent {
     }))
   }
 
+
   render() {
-    let testCoverageMaxima, networkCoverageMaxima
+    const { testCoverage, networkCoverage } = this.props
+    
+    // Check if there is enough data to plot the charts
+    const testCoverageCount = testCoverage.reduce((count, item) => count + item.count, 0)
+    const networkCoverageCount = networkCoverage.reduce((count, item) => count + item.count, 0)
+    const notEnoughData = (testCoverageCount === 0 && networkCoverageCount === 0)
 
-    const { testCoverage, networkCoverage, notEnoughData } = this.props
     const supportedTestGroups = ['websites', 'im', 'middlebox', 'performance', 'circumvention']
-    const selectedTestGroups = Object.keys(this.state).filter(testGroup => this.state[testGroup])
 
-    const testCoverageByDay = testCoverage.reduce((prev, cur) => {
-      prev[cur.test_day] = prev[cur.test_day] || {}
-      if (selectedTestGroups.indexOf(cur.test_group) > -1) {
-        prev[cur.test_day][cur.test_group] = cur.count
-        const allTestCount = Object.values(prev[cur.test_day]).reduce((p,c) => p+c, 0)
-        if (typeof testCoverageMaxima === 'undefined'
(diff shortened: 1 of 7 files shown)


components/country/overview.js
```diff
@@ -139,6 +141,7 @@ const Overview = ({
         <FormattedMessage id='Country.Overview.Heading.TestsByClass' />
       </Heading>
       <TestsByGroup
+        notEnoughData={measurementCount < LOW_DATA_THRESHOLD}
```
sarathms: @hellais Please review if this is a good measure to detect when to show 'No Data Available' message across the overview chart. For reference, the `/country/IO` page has 4 measurements here, but renders an empty chart.

hellais: What you really want to be looking at is if there is data from the last 30 days.

hellais: You should not be looking at the absolute measurementCount, but rather the `testCoverage` and `networkCoverage` arrays.

A future optimisation could be to have an API endpoint that is just for this (or maybe do it as part of other API endpoints).

In any case it's not enough to look at the absolute `measurementCount`, because if a country has many measurements, but they are all in the past we still don't have enough data to plot the overview chart.

hellais: We may also want to change the copy to "No data available from the last 60 days" cc @agrabeli

hellais:
You should also see if it's easy to ensure that while the chart is still loading you don't say "No data available". I noticed that some of the charts in the lower part of the page say "No data avialable" at the beginning of the load, which may not always be as fast.

I think it's also fine if it's too much work to document this as an issue and defer to future work.

components/country/overview.js
sarathms: :+1: done

components/country/overview-charts.js
```diff
@@ -72,42 +72,154 @@ class TestsByGroup extends React.PureComponent {
...
   }
 
+
   render() {
-    let testCoverageMaxima, networkCoverageMaxima
+    const { testCoverage, networkCoverage } = this.props
+    
+    // Check if there is enough data to plot the charts
+    const testCoverageCount = testCoverage.reduce((count, item) => count + item.count, 0)
+    const networkCoverageCount = networkCoverage.reduce((count, item) => count + item.count, 0)
```
hellais: This is maybe not so efficient and we should maybe consider adding an API endpoint to serve this need.

I documented this here: https://github.com/ooni/api/issues/99

hellais:
I think there is something odd in how it's handled in the websites section:
Screenshot 2019-08-20 at 10 41 23

Here are some things to change:

* It says 1 page of 0 pages (you should disable the pagination entirely)
* The ASN selector is still active and some of the measurements there have a count of measurements > 0
* The text for no data available should be bigger

sarathms:
> I noticed that some of the charts in the lower part of the page say "No data avialable" at the beginning of the load,

I couldn't see this happening. Anyway, I have opened #264 for this.

sarathms:
> * It says 1 page of 0 pages (you should disable the pagination entirely)
> * The ASN selector is still active and some of the measurements there have a count of measurements > 0

Have been addressed in #263

Issue #250, opened 2019-08-13
Implement lazy loading for charts


holantonela:
As part of #173, we discussed how to improve users perception of loading when our heavy charts are loading. @hellais mentioned that https://github.com/danilowoz/react-content-loader is already in use.

hellais:
This is now implemented in the latest version of Explorer.

PR #263, opened 2019-08-19
Better loading indicators

### LINKED ISSUE #243: Replace search page loader with content loader
Implement a content loader similar to the one used for landing page chart. Using: [`create-content-loader`](http://danilowoz.com/create-content-loader/).

sarathms:
Related to #250 
Closes #243


components/search/loader.js
```diff
@@ -1,35 +1,46 @@
...
+    <rect x="160" y="180" rx="3" ry="3" width="70" height="5" />
+    <rect x="250" y="180" rx="3" ry="3" width="70" height="5" />
+    <circle cx="20" cy="213" r="5" />
+    <rect x="40" y="210" rx="3" ry="3" width="100" height="5" />
+    <rect x="160" y="210" rx="3" ry="3" width="70" height="5" />
+    <rect x="250" y="210" rx="3" ry="3" width="70" height="5" />
+    <circle cx="20" cy="243" r="5" />
+    <rect x="40" y="240" rx="3" ry="3" width="100" height="5" />
+    <rect x="160" y="240" rx="3" ry="3" width="70" height="5" />
+    <rect x="250" y="240" rx="3" ry="3" width="70" height="5" />
```
hellais: Maybe the loader svgs should go in a specific svg file so they can be easily updated and modified by designers?

hellais: Actually I see that it's not really a simple svg. I will push some changes to this directly on the branch.

sarathms:
I added loaders for the website charts to address the jumping problem. If it looks good, we can add the same for IM charts too.

hellais:
This looks good, however it still "jumps" when you first go to the next page and it renders nothing.

I will share a screencap on slack.

sarathms:
Added some more fixes. The entire section doesn't collapse anymore. But the loaders move a few pixels as the charts start rendering. It doesn't happen in the 'Apps' section though, the switch from loader to actual chart is pretty smooth.

components/country/websites-charts.js
```diff
@@ -139,16 +139,19 @@ class TestsByCategoryInNetwork extends React.Component {
           </FormattedMessage>
         */}
         {/* URL-wise barcharts Start */}
-        {fetching && renderLoader()}
-        {(!fetching && testedUrls) &&
+        <Box style={{'min-height': '500px'}}>
```
hellais: It maybe doesn't actually have to be `Box` it can just be a `div`

sarathms: Does it fix the issue though? I still see jitters as the page renders more stuff. Moreover, it creates another problem to solve on mobile and other layouts, which is why I chose not to go that route. It seems to work perfectly in the `Apps` section without fixing the container heights.

hellais: Good point about the mobile views, I had not considered that.

Locally and on desktop this did fix the jitter issues for me. Is this branch deployed to explorer-beta?

hellais: In second thought I think you should revert my commit and keep it how it is now and document this in an issue for future work.

sarathms: Opened #278 to address this later.

hellais:
This looks good to me. I also can't reliably reproduce the jumping anymore in my browser. 🐳

PR #265, opened 2019-08-22
Add input validations

### LINKED ISSUE #12: Implement the search page

sarathms:
Closes #12 

Validates `ASN` and `Input` fields. Shows error message below the field.
image

image

### DIFF
--- a/components/search/filter-sidebar.js
+++ b/components/search/filter-sidebar.js
@@ -64,7 +64,6 @@ class FilterSidebar extends React.Component {
       showUntilCalendar: false,
       isFilterDirty: false,
       asnError: false,
-      inputError: false,
     }
     this.onChangeFilter = this.onChangeFilter.bind(this)
     this.onDateChangeFilter = this.onDateChangeFilter.bind(this)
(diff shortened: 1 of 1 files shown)


components/search/filter-sidebar.js
```diff
@@ -73,6 +75,45 @@ class FilterSidebar extends React.Component {
...
+            asnError: false
+          })
+        }
+        break
+
+      case 'inputFilter':
+        var inputValue = e.target.value
+        // Regex from: https://www.regexpal.com/93652
+        // eslint-disable-next-line no-useless-escape
+        var inputRegex = /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?$/
```
hellais: As discussed, we should change this to use a domain name and only check if it's a valid domain name

PR #279, opened 2019-08-30
Use domain to search

### LINKED ISSUE #274: Replace "input" with "Domain" in Search page
Edit the copy to replace "Input" with "Domain" in the Search page so it's clear to users what we mean.

### LINKED ISSUE #281: Fix filter for tested URLs in country page
When exploring the measurements on the Websites section on the ET country page, I clicked on "Explore more measurements" under a tested URL (http://africanarguments.org/). I expected to get re-directed to the Search page and only view measurements filtered based on that URL.

Instead, I was re-directed to the Search page including what appears to be a random list of measured URLs (not filtered based on http://africanarguments.org/). See: https://explorer-beta.ooni.io/search?test_name=web_connectivity&probe_cc=ET&probe_asn=24757&input=http%3A%2F%2Fafricanarguments.org&since=2019-08-03&until=2019-09-02

random-urls

This probably needs to be fixed so that the user is re-directed to the Search page containing only measurements pertaining to the selected URL.

sarathms:
Closes #274 
Closes #281 

Uses the `domain` parameter for the search request. Validates text field to accept only valid domains in the following formats:
```
google.com
https://google.com
http://google.com
www.google.com
www.google.com/
google.com:8080/
```


components/search/filter-sidebar.js
```diff
@@ -119,21 +120,40 @@ class FilterSidebar extends React.Component {
...
-            asnError: false
+            asnError: false,
+            isFilterDirty: true
           })
         }
         break
+      case 'domainFilter':
+        var domainValue = e.target.value
+        // eslint-disable-next-line no-useless-escape
+        var domainRegEx = /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/)?$/
```
hellais: The `http://` or `https://` prefix must not be present for it to be a valid domain name.

hellais:
The example in the search sidebar should show just a domain, `ex. twitter.com`.

components/search/filter-sidebar.js
sarathms: True. Can we not accept values with valid `http://` or `https://` and truncate them before passing on to the API? I am guessing people might copy paste such values often and end up thinking there is no data about them.

hellais: As we are changing the behaviour to accepting domains, we should rather show a validation error if the user does not enter a valid domain and maybe in the validation error mention that it should not have the `http` prefix

sarathms: The validation error component doesn't work very well with longer messages. It already looks unpleasant this way.
image

We should be able to handle this better when we do https://github.com/ooni/design-system/issues/62

sarathms: Updated regex to validate for domain names without prefixes.

sarathms:
> example in the search sidebar should show just a domain

Updated copy accordingly.

hellais:
Can we add back the `e.g.` prefix before the domain? For the rest this looks good to merge to me.

sarathms:
This PR now closes #281 also.

sarathms:
> Can we add back the e.g. prefix

It is part of the copy now. I have updated the copy. Haven't deployed it though.

PR #284, opened 2019-09-06
Move country summary to top of page

### LINKED ISSUE #270: Move overview paragraph to top header of Country pages
Based on UX feedback collected at CCCamp, it was proposed that we move the overview paragraph in the Country pages to the top header, right under the Overview heading.

### LINKED ISSUE #271: Edit copy of Overview paragraph to trim it down
Based on UX feedback, it was proposed that the copy of the Overview paragraph is made more concise. Perhaps consider replacing the paragraph with bullet points, or with a single sentence, or widgets.

For example:

Since XXX OONI Probe users have collected YYY measurements from ZZZ local networks.

Explore the data below to check the accessibility and/or blocking of sites and services.

sarathms:
Closes #270 
Closes #271


sarathms:
I deployed it for now because I wanted to fix the missing translation key for the country page summary text paragraph.

It still doesn't have the
```
and highlight the bits of text which are variable
```
part of the issue yet.

sarathms:
@hellais The simplest way to achieve highlighting the variable parts of the paragraph, is by doing it in markdown. The entry in the copy is now like this:
```
"Since **{startDate}**, OONI Probe users in **{countryName}** have collected **{measurementCount}** measurements from **{networkCovered}** local networks."
```

I went ahead with this because we expect this paragraph to be markdown formatted anyway. We have so far used it only as a way to easily include hyperlinks in translated texts. But I figured it is not any more complicated to bold a few parts of the paragraph.

hellais:
LGTM

PR #286, opened 2019-09-06
Rearrange fields in search filter

### LINKED ISSUE #275: Swap the order of Test Name & Country drop-down menus in Search page
Community members proposed that we swap the order of the Test Name & Country drop-down menus in the Search page so that the Country drop-down menu is presented first.

This would be an important thing to do, particularly since most users probably won't recognize OONI Probe test names.

sarathms:
Closes #275


hellais:
LGTM

PR #293, opened 2019-09-11
Use UTC mode when parsing and displaying time

sarathms:


components/search/results-list.js
```diff
@@ -305,8 +305,8 @@ const ResultItem = ({msmt}) => {
                 <Box width={3/16}>
                   <ASNBox asn={msmt.probe_asn} />
                 </Box>
-                <Box width={4/16}>
-                  {moment(msmt.measurement_start_time).format('YYYY-MM-DD')}
+                <Box width={4/16} title={moment.utc(msmt.measurement_start_time).format()}>
+                  {moment.utc(msmt.measurement_start_time).format('YYYY-MM-DD')}
```
hellais: This needs to include the time as well.

So it should be:
```
{moment.utc(msmt.measurement_start_time).format('HH:mm YYYY-MM-DD')
```

sarathms: I show the time in a tooltip. Including time in the table made it look quite cluttered when I tried yesterday. And I didn't want to go about adjusting the row layout at this point. But then I used `YYYY-MM-DD hh:mm a` format. If we use the 24-hour format like you suggested, it doesn't look that bad.

sarathms: I have retained the tooltip to be able to include `UTC` in it to help clarify that for users who wonder.

image

PR #323, opened 2019-09-25
Improve error reporting on failed searches

### LINKED ISSUE #321: Error after receive timeout error (json) from API: "There was an unexpected error from our end."
## Expected Behavior
Show timeout error message

## Actual Behavior
Page crashed and displays an error:
```
There was an unexpected error from our end.
Maybe try exploring some measurements or go to our homepage.
```

  1.  open: https://explorer.ooni.org/search?until=2019-09-18&domain=anidalton.blogfa.com&only=confirmed

## API response 
```json
{
  "detail": [
    "The database query timed out.",
    "Try changing the query parameters."
  ],
  "status": 504,
  "title": "Gateway Timeout",
  "type": "about:blank"
}
```

sarathms:
Shows an error message with details when search query times out.

Closes #321 

image

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -155,10 +155,12 @@ class Search extends React.Component {
 
     try {
       msmtR = await getMeasurements(query)
-    } catch (error) {
-      // Remove `error.response.request` which has a circular reference to `response`
-      if (error.response && error.response.request) {
-        delete error.response['request']
+    } catch (err) {
+      let error
+      if (err.response) {
+        error = err.response.data
+      } else {
+        error = err.message
       }
       return {
         error,
--- a/pages/search.js
+++ b/pages/search.js
@@ -172,6 +172,7 @@ class Search extends React.Component {
     const measurements = msmtR.data
 
     return {
+      error: null,
       results: measurements.results,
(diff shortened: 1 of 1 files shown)


pages/search.js
```diff
@@ -88,8 +89,23 @@ const ErrorBox = ({ error }) => {
...
+        </Box>
+      </Flex>
+      <Flex justifyContent='center'>
+        <Box>
+          <Heading h={4} my={4}>
+            We encountered an error. Please try again or change the filters to get different results.
+          </Heading>
+          <Box p={[1, 3]} bg='gray3'>
+            <pre>
+              {JSON.stringify(error, null, '  ')}
```
hellais: We don't actually need the full error object, we just need `error.response.data`.

hellais: Perhaps it's useful to also include the statusCode returned by the API backend. 

So something like:

```
{
JSON.stringify(
{
"data": error.response.data,
"statusCode": error.response.statusCode
}
, null, '  ')
}
```

sarathms: Shows this now
image

PR #327, opened 2019-09-29
Use correct icon for Vanilla Tor in circumvention section

sarathms:
Upgrades `ooni-components` to use the right icon like in the screenshot.

image


hellais:
lgtm

PR #328, opened 2019-10-01
Adds UTC after displayed time in measurement page

### LINKED ISSUE #305: Add UTC in measurement pages
Currently it's unclear in measurement pages whether the referred time of testing is in the local timezone or in UTC. In fact, it's most likely to be read as being in the local timezone.

I therefore propose that UTC is displayed along with the time of testing in the measurement pages.

For example: On Sep 5, 2019 12:32 PM *UTC*, https://www.tunnelbear.com/ was accessible when tested on AS9500 in New Zealand (taken from: https://explorer.ooni.io/measurement/20190906T215018Z_AS9500_eO6TdiWobglEx2RmWNFmKqtrPDFaIDTGsJxeypOuuhkQZ69jlp?input=https://www.tunnelbear.com/).

SaptakS:
Fixes #305 

I have added "UTC" to the datetime. Added via intl, just to be safe and to track in i18n.

Following is how it looks:
Screenshot-2019-10-1 OONI Explorer - Open Data on Internet Censorship Worldwide


components/measurement/CommonSummary.js
```diff
@@ -78,6 +79,7 @@ const CommonSummary = ({
             />
             <SummaryItemBox
               label={intl.formatMessage({ id: 'Measurement.CommonSummary.Label.DateTime' })}
+              timezone={intl.formatMessage({ id: 'Measurement.CommonSummary.Label.UTC' })}
```
sarathms: @SaptakS Maybe we can keep `SummaryItemBox` a generic component and include the `timezone` label in `content` itself, like it was done for [`countryBlock`](https://github.com/ooni/explorer/pull/328/files#diff-7d296e25cd6dae8a5f0748bec6003226R74). That way `SummaryItemBox` need not know about `timezone` as a prop and we can also avoid having to do a propType check for it.

hellais:
> Do you have any comments about the way "UTC" is displayed here as per screenshot? I am fine with it.

It looks good to me. Are we wrapping the date with some intl related function so the formatting of the time is in the locale of the user?

If we are not doing that maybe it would be wise to use the more standard 24h format for displaying time.

sarathms:
> Are we wrapping the date with some intl related function
It is now formatted with `intl.formatDate` from `react-intl` which picks up the locale from its context.

PR #329, opened 2019-10-01
Adds ooni version from package.json to footer

### LINKED ISSUE #317: Add OONI Explorer version number to footer
This could come from `package.json`.

SaptakS:
Fixes #317 

Imports version from package.json and adds it to the footer below the logo.

Following is how it looks:
Screenshot-2019-10-1 OONI Explorer - Open Data on Internet Censorship Worldwide(1)


components/footer.js
```diff
@@ -51,6 +52,11 @@ const FooterText = styled.div`
   margin-top: 0px;
 `
 
+const FooterVersion = styled.div`
+  margin-top: 0px;
+  margin-left: 45px;
+`
+
```
sarathms: If we are only customizing the margins, I think it is better to just use the `Box` component we already use. [This documentation](https://rebass-v2.now.sh/props#margin-and-padding) should help. 

It might also help with the overall consistency if we use the [scale set in the theme](https://github.com/ooni/design-system/blob/master/components/theme/rebassTheme.js#L10-L16) e.g `ml={[1-6]}` instead of a fine tuned value like `45px`

components/footer.js
```diff
@@ -59,6 +65,7 @@ const Footer = ({ intl }) => (
           <Flex flexWrap='wrap' alignItems='center'>
             <Box p={[1, 0]} mb={[0, 3]} width={[1/2, 1]}>
               <ExplorerLogo height='32px' />
+              <small><FooterVersion>{intl.formatMessage({ id: 'Footer.Text.Version' }) }: {version}</FooterVersion></small>
```
sarathms: It might be a bit confusing to put the version number right under 'OONI' which represents the probe or the entire ecosystem of tools, and the version number itself is of the 'Explorer'. We can show the version number in a further less prominent spot. You can choose from:
* right under the Creative Commons license notice
* on the far right end of the copyright and license notice, as a sibling node to [this `<FooterBox>`](https://github.com/ooni/explorer/blob/master/components/footer.js#L89). You can add `justifyContent='space-between'` to the containing `<Flex>`

components/footer.js
SaptakS: That makes total sense. I was trying to keep it consistent with `FooterText` but didn't know about the scale. Thanks for pointing me in the right direction.

components/footer.js
SaptakS: Thanks for pointing this out. Makes sense. Going to fix it now.

SaptakS: I think it looks better under the Creative Commons rather than far end since it kind of clashes with the feedback icon.

Following are the screenshots:

Under:
Screenshot-2019-10-2 OONI Explorer - Open Data on Internet Censorship Worldwide(1)

Far right end:
Screenshot-2019-10-2 OONI Explorer - Open Data on Internet Censorship Worldwide

SaptakS: Also, I was wondering if `OONI Explorer Version: ` would be a a better text than just `Version` to make it clear.

sarathms: Please go ahead with putting it under the license notice.

> wondering if OONI Explorer Version: would be a a better text 
You can leave it as 'Version' for now. The rendered text will be generated from a spreadsheet which we coordinate with our translators. The English version will also be edited by someone. Thanks for considering.

SaptakS:
@sarathms Made the necessary changes as requested by you.

components/footer.js
```diff
@@ -85,11 +86,12 @@ const Footer = ({ intl }) => (
           <FooterLink href='https://openobservatory.slack.com/' label={intl.formatMessage({ id: 'Footer.Link.Slack' })} />
         </FooterBox>
       </Flex>
-      <Flex flexWrap='wrap'>
+      <Flex flexWrap='wrap' justifyContent='space-between'>
```
sarathms: You can drop this change. It is useful only when there are multiple child nodes inside `<Flex>`. This is my [go-to link](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) for all things flex.

SaptakS:
I force pushed because I didn't want there to be two separate commits for small change. As far as I know, we should have commits as squashed and managed as possible in a PR. And if I rebase in git, which we have to force push.

components/footer.js
SaptakS: I forgot to remove it after experimenting with the position of the Version. That's all. Not like I left it on purpose.

SaptakS:
Done as requested. Made a separate commit for code removal (as suggested, instead of rebasing and squashing)

sarathms:
> I force pushed because I didn't want there to be two separate commits for small change. As far as I know, we should have commits as squashed and managed as possible in a PR. And if I rebase in git, which we have to force push.

Agreed. That is a valid way of doing things as well.

Thanks for your contribution. Sorry if I took too long to respond.

PR #342, opened 2019-11-13
Add links to RSS feeds from Fastpath event detector

### LINKED ISSUE #330: Add RSS feed to publish updates about detected censorship events
When https://github.com/ooni/spec/issues/144 is done, we should be able to publish an RSS feed which user's can subscribe to receive updates about censorship events detected in our pipeline.

sarathms:
This provides RSS feeds to subscribe to events detected by 'fastpath' (a component in the pipeline). The following different feeds will be made available:
- [x] A global feed added too all pages
- [x] A country wide feed added to the country pages

Fixes #330

### DIFF
--- a/components/header.js
+++ b/components/header.js
@@ -13,6 +13,7 @@ export default class Header extends React.Component {
           <meta charSet='utf-8'/>
           <meta name='viewport' content='initial-scale=1.0, width=device-width'/>
           <meta name='description' content={description} />
+          <link rel="alternate" title="Events Detected by OONI" href="http://fastpath.ooni.nu:8080/rss/global.xml" type="application/rss+xml" />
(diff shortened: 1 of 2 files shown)


components/header.js
```diff
@@ -13,6 +13,7 @@ export default class Header extends React.Component {
           <meta charSet='utf-8'/>
           <meta name='viewport' content='initial-scale=1.0, width=device-width'/>
           <meta name='description' content={description} />
+          <link rel="alternate" title="Events Detected by OONI" href="http://fastpath.ooni.nu:8080/rss/global.xml" type="application/rss+xml" />
```
hellais: This should rather be: https://explorer.ooni.org/rss/global.xml

sarathms: This PR wasn't ready for review yet :grin:

PR #353, opened 2019-12-02
Fix multiple API requests on search page

### LINKED ISSUE #350: API requests launched twice on client side
State changes in components wired to API requests seem to trigger duplicate requests. This needs to be investigated and fixed if relevant. For example, the search form.

sarathms:
To update the URL as we update search filters, we were using `Router.push` that triggered navigation, which in turn called `getInitialProps` and the API requests right after they were already initiated in `onApplyChange`.

The solution according to this [spectrum-chat](https://spectrum.chat/next-js/general/change-url-without-triggering-navigation~12ba234c-ad05-4783-8342-8a4a07661a9c) is to perform a shallow ~rendering~ routing which updates the URL without navigation.

Fixes #350

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -208,6 +208,15 @@ class Search extends React.Component {
     this.loadMore = this.loadMore.bind(this)
   }
 
+  componentDidMount () {
+    const { query, replace } = this.props.router
+    const href = {
+      pathname: '/search',
+      query
+    }
+    replace(href, href, { shallow: true })
+  }
+
   shouldComponentUpdate (nextProps, nextState) {
     if (this.props != nextProps) {
       return true
--- a/pages/search.js
+++ b/pages/search.js
@@ -201,22 +201,11 @@ class Search extends React.Component {
       search: null,
       error: props.error,
 
-      loading: true
+      loading: false
     }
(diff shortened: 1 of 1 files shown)


sarathms:
@hellais Did you do that in `dev` mode? I am assuming `next.js` handles this in fine production (the file names are hashed right?)

hellais:
> Did you do that in dev mode?

I did all the tests using something close to what is done in production, that is running:
```
yarn run build && yarn run start
```

>  I am assuming next.js handles this in fine production (the file names are hashed right?)

That is what I was assuming too, which is why I was confused that I had to run to a force reload in order to see the changes take effect. Maybe I did something wrong, I just figured I would bring it up as maybe you have looked into this already.

sarathms:
I don't think this is problem enough to do something right now. There seem to be more factors involved. Lets keep an eye out. Meanwhile if you are able to reproduce it, lets open an issue for it separately. Going ahead with merge now.

sarathms:
A note about navigating out and back into the search page: I think we should consider caching the API responses locally. I am looking at a story where a user searches, gets a bunch of results and wants to go through them one by one. Unless the use opens each row in a separate tab, a full search query happens every time we go back to the screen.

PR #364, opened 2019-12-27
Fixes the until today logic to include measurements today

### LINKED ISSUE #361: Default value of until filter should include measurements of today
Currently the `until` filter is set to default to the current date. The problem with this is that I would expect it to filter inclusive of all measurements for the date of today, but instead it filters measurements until midnight of today (effectively yesterday).

I think we should fix this by making it so the until filter is set by default to the date of tomorrow instead of the date of today.

SaptakS:
Fixes #361 

I have changed the `today` value everywhere to add 1 day so that the filter includes today's measurements. In doing so, I have changed the name of the variable from `today` to `until` which is something I saw used in the API repository. `today` sounded like a bad variable name, since clearly it's `tomorrow`. I am up to changing the variable name from `until` to something like `tomorrow` or `untilToday` if you all feel that's more suitable/relevant.

### DIFF
--- a/components/search/FilterSidebar.js
+++ b/components/search/FilterSidebar.js
@@ -50,7 +50,7 @@ class FilterSidebar extends React.Component {
     super(props)
 
     // Display `${today}` as the end date for default search
-    const today = moment().format('YYYY-MM-DD')
+    const until = moment().add(1, 'day').format('YYYY-MM-DD')
 
     this.state = {
       domainFilter: props.domainFilter || '',
@@ -59,7 +59,7 @@ class FilterSidebar extends React.Component {
       countryFilter: props.countryFilter || '',
       asnFilter: props.asnFilter || '',
       sinceFilter: props.sinceFilter || '',
-      untilFilter: props.untilFilter || today,
+      untilFilter: props.untilFilter || until,
       showSinceCalendar: true,
       showUntilCalendar: false,
       isFilterDirty: false,
--- a/components/search/FilterSidebar.js
+++ b/components/search/FilterSidebar.js
@@ -49,7 +49,8 @@ class FilterSidebar extends React.Component {
   constructor(props) {
     super(props)
 
-    // Display `${today}` as the end date for default search
+    // Display `${tomorrow}` as the end date for default search
+    // to include the measurements of `${today}` as well.
     const until = moment().add(1, 'day').format('YYYY-MM-DD')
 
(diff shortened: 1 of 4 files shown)


components/search/FilterSidebar.js
```diff
@@ -50,7 +50,7 @@ class FilterSidebar extends React.Component {
     super(props)
 
     // Display `${today}` as the end date for default search
```
bassosimone: Can you also update this comment?

pages/search.js
```diff
@@ -133,9 +133,9 @@ class Search extends React.Component {
     // By default, on '/search' show measurements published until today
     // This prevents the search page from showing time-travelling future
     // measurements from showing up
```
bassosimone: Can you also update this comment to more precisely explain the new situation?

SaptakS:
@bassosimone changed the comments to convey the message. Please let me know if the wording sounds right to you or you would prefer something else.

components/search/FilterSidebar.js
```diff
@@ -49,8 +49,9 @@ class FilterSidebar extends React.Component {
   constructor(props) {
     super(props)
 
-    // Display `${today}` as the end date for default search
-    const today = moment().format('YYYY-MM-DD')
+    // Display `${tomorrow}` as the end date for default search
+    // to include the measurements of `${today}` as well.
+    const until = moment().add(1, 'day').format('YYYY-MM-DD')
```
hellais: Is this `until` date going to be the current date in UTC + 1?

It's worth considering that the `until` filter in the API uses the time in UTC, while I suspect that `moment()` will by default use the timezone of the browser.

We should take that into account and here compute 1 day from now in UTC.

bassosimone: This is a very good point! So, we need to use `moment().utc()` according to https://momentjs.com/docs/#/parsing/utc/.

SaptakS: This is a good point. I will send an updated PR for the same. Also, this makes me think, shouldn't all the date (like `since`) be also in UTC. Otherwise that also would give based on timezone.

bassosimone: Yes, I believe there is a bunch of places where we're not using UTC and we should probably make sure we're using the proper time in the proper context.

I filed https://github.com/ooni/explorer/issues/368 to track this localtime->UTC work.

PR #367, opened 2019-12-29
Adds regex to allow IP search in domain filter

### LINKED ISSUE #354: Ability to search for IP
These four URLs cannot be searched in explorer:
https://github.com/citizenlab/test-lists/blob/master/lists/global.csv#L1148-L1151

The problem is from the frontend.
http://explorer.ooni.org/search?until=2019-12-04&domain=1.1.1.1&test_name=web_connectivity&since=2019-11-01

## Expected Behavior
Ability to search for IP

## Actual Behavior
error: `Please enter a valid domain name, such as twitter.com`

  1. write: 1.1.1.1

SaptakS:
Should fix #354 

Things done in this PR:
- Regex checks to allow for IP addresses in the domain filter in the sidebar. I have made a separate regex pattern for IP, just in case we want to create a separate regex only for IP or a separate UI filter or use only the IP regex somewhere else. However if the general consensus is to add the regex in the current domain regex, then I am flexible to do that.
-  I have also increased the TLD part of regex in domain filter from 5 to 7. This is to accomodateTLDs like `website` which i myself use for my personal website.
- Also updated the error message and placeholder to better give idea.


hellais:
I am not 100% sure the changes we are rolling out in the API which optimise the query time, will work with IP addresses.
@FedericoCeratto do you know?

static/lang/en.json
```diff
@@ -298,8 +298,8 @@
   "Country.NetworkProperties.InfoBox.Label.Middleboxes.NotFound": "No middleboxes detected",
   "Country.NetworkProperties.Button.ShowMore": "Show more networks",
   "Search.Sidebar.Domain": "Domain",
-  "Search.Sidebar.Domain.Placeholder": "e.g. twitter.com",
-  "Search.Sidebar.Domain.Error": "Please enter a valid domain name, such as twitter.com",
+  "Search.Sidebar.Domain.Placeholder": "e.g. twitter.com or 1.1.1.1",
+  "Search.Sidebar.Domain.Error": "Please enter a valid domain name or IP address, such as twitter.com or 1.1.1.1",
```
hellais: We should remember to update the master translations file with the updated strings. cc @sarathms

SaptakS: @hellais isn't the `static/lang/translations.js` the master translations file? Because I have updated that.

hellais: We also have an internal spreadsheet which is used to generate all the translation files.

SaptakS: okay

PR #373, opened 2020-01-16
Enables UTC timings in codebase

### LINKED ISSUE #368: Make sure we use UTC time when showing measurements
Following up from https://github.com/ooni/explorer/pull/364#pullrequestreview-337165238, we should make sure we're using UTC time when showing measurements. I am not so familiar with the codebase to state that we must always use UTC, but we most likely need UTC in most cases.

(It _may_ be a nice feature to _also_ show the presumed local time when the measurement was taken based on the country, but this strikes me as an entirely different feature.)

SaptakS:
Fixes #368


components/search/FilterSidebar.js
```diff
@@ -51,7 +51,7 @@ class FilterSidebar extends React.Component {
 
     // Display `${tomorrow}` as the end date for default search
     // to include the measurements of `${today}` as well.
-    const until = moment().add(1, 'day').format('YYYY-MM-DD')
+    const until = moment().utc().add(1, 'day').format('YYYY-MM-DD')
```
hellais: It's not introduced in this PR, but I maybe we want to fix it now too.

The until filter does not allow setting the date of tomorrow, although it's the default which is selected.

See:
Screenshot 2020-01-17 at 12 38 25

components/country/AppsStatsCircumventionRow.js
```diff
@@ -149,7 +149,7 @@ class AppsStatsCircumventionRow extends React.Component {
               <Box ml='auto'>
                 <FormattedMessage id='Country.Apps.Label.LastTested' />
                 {' '}
-                <strong>{moment(data.last_tested).fromNow()}</strong>
+                <strong>{moment.utc(data.last_tested).fromNow()}</strong>
```
hellais: Maybe we want to add after here the string UTC to clarify the timezone.

hellais:

I added a few minor things to look into. 

For the fixing of the until filter, it's also fine if we just document that as a separate issue and do it as part of another part of work.

components/search/FilterSidebar.js
SaptakS: I think this might have to do something with the calendar part of the component which allows till today as a maximum date. If you don't mind, maybe can handle this as a separate issue?

components/country/AppsStatsCircumventionRow.js
SaptakS: Makes sense. I will update similarly in other places too

components/search/FilterSidebar.js
hellais: Sounds good. I filed this issue for it: https://github.com/ooni/explorer/issues/375

components/country/AppsStatsCircumventionRow.js
SaptakS: Done!

hellais: Ah sorry I think I did not explain myself properly. What I meant is that you should add after
```
<strong>{moment.utc(data.last_tested).fromNow()}</strong>
```

Literally the string `UTC` so that the user understands that the time is being displayed in the UTC timezone.

However I see now that actually you are using `fromNow()` which according to the docs says: https://momentjs.com/docs/#/displaying/fromnow/
```
moment([2007, 0, 29]).fromNow();     // 4 years ago
```

So it doesn't actually make sense to do that here.

I think what you want to be doing is actually revert the commit https://github.com/ooni/explorer/pull/373/commits/77a3fb735dfadd61a5acbf48e1016fc0628e9e2d.

We should however verify if the delta of time is correct, but I believe it should be, because by doing `moment().utc(last_tested)` we are telling moment that the timezone of the source string is UTC and it should calculate the delta in accordance to that.

Sorry for the confusion.

SaptakS: Ooooh. Okay. Got it.

SaptakS: So what I will do is add the string UTC in places where it makes sense. As in where it's actually a date and not delta.

SaptakS: @sarathms @hellais I think I should approach it like https://github.com/ooni/explorer/pull/328/commits/e0ee6147a530ec1f09c95807d88c74ffc2ad55b8 instead of just adding a text? Because then it will be similar.

SaptakS:
I have added the necessary timezone text after dates following the same method as in https://github.com/ooni/explorer/pull/328/commits/e0ee6147a530ec1f09c95807d88c74ffc2ad55b8

components/measurement/SummaryText.js
```diff
@@ -14,8 +14,16 @@ const SummaryText = ({
...
+  const formattedDate = moment(date).format('LL')
+  const formattedDateTime = intl.formatDate(moment.utc(date).toDate(), {
+    year: 'numeric',
+    month: 'long',
+    day: '2-digit',
+    hour: '2-digit',
+    minute: 'numeric',
+    timeZone: 'UTC',
+    timeZoneName: 'short'
+  })
```
hellais: @sarathms is this way of displaying dates consistent with how we do it in other places?

components/search/ResultsList.js
```diff
@@ -306,7 +306,7 @@ const ResultItem = ({msmt}) => {
                   <ASNBox asn={msmt.probe_asn} />
                 </Box>
                 <Box width={4/16} title={moment.utc(msmt.measurement_start_time).format('YYYY-MM-DD HH:mm [UTC]')}>
-                  {moment.utc(msmt.measurement_start_time).format('YYYY-MM-DD HH:mm')}
+                  {moment.utc(msmt.measurement_start_time).format('YYYY-MM-DD HH:mm [UTC]')}
```
hellais: Same for this, we should check if the `UTC` string goes inside of square brackets or not.

SaptakS: I did check and seems inside bracket works for me. But @sarathms do let me know if you think this is not a good way.

components/measurement/SummaryText.js
SaptakS: This actually follows the way @sarathms did in https://github.com/ooni/explorer/commit/e0ee6147a530ec1f09c95807d88c74ffc2ad55b8. I followed that

components/search/ResultsList.js
sarathms: 1. For consistency,  let us use `intl.formatDate` for places where we display dates.

2. I am not sure we have real estate to add the UTC string to the date column here. It looks pretty cluttered. This is why we made a compromise and added a tooltip which showed the same value but with the UTC suffix.

image

Unless there is a better UX solution, I suggest we revert this show what it showed earlier, but using the `intl.formatDate()` API instead of `moment.format()` for consistency.

pages/search.js
```diff
@@ -139,7 +139,7 @@ class Search extends React.Component {
     // including the measurements of today (so the date of tomorrow).
     // This prevents the search page from showing time-travelling future
     // measurements from showing up
-    const until = moment().add(1, 'day').format('YYYY-MM-DD')
+    const until = moment().utc().add(1, 'day').format('YYYY-MM-DD')
```
sarathms: `moment.utc()`

SaptakS:
@sarathms added changes based on your comments

components/search/ResultsList.js
```diff
@@ -306,7 +306,13 @@ const ResultItem = ({msmt}) => {
                   <ASNBox asn={msmt.probe_asn} />
                 </Box>
                 <Box width={4/16} title={moment.utc(msmt.measurement_start_time).format('YYYY-MM-DD HH:mm [UTC]')}>
-                  {moment.utc(msmt.measurement_start_time).format('YYYY-MM-DD HH:mm')}
+                  {intl.formatDate(moment.utc(msmt.measurement_start_time).toDate(), {
```
sarathms: You need to call the `useIntl` hook inside the FC to be able to use `intl` here.

SaptakS: Sorry.  I misunderstood in the previous comment.

SaptakS:
@sarathms updated the code based on your request. I have used `useIntl()` only in the new codes added in this PR. There seems to be a lot of places where we are using the `injectIntl`. Let me know if you want me to update all those

SaptakS:
@sarathms, I am happy to do that though might need a bit time since I am travelling. Won't that upgrade be kind of blocker for showing times in UTC? If it's not much priority then I am fine with going down that route.

sarathms:
No worriesl You can continue working on this whenever you can.

SaptakS:
@sarathms Just made a commit to upgrade react-intl and replace `injectIntl` HOC with `useIntl` hooks in function components. There is still `injectIntl` being used in `FilterSidebar.js`, `OverviewCharts.js` and `Stats.js`  since they are class components and hooks don't work in class components.

hellais:
Thanks for the heads up, can we preserve the formatting of the timestamp as it was previously?

I think that if we are going to go with a date formatting which includes the names of months and is displayed different this should be done in a separate PR and after thinking a bit more about the UX of this solution.

SaptakS:
@hellais we can maintain the formatting to something like:
image

But the format is MM/DD/YYYY instead of the current YYYY-MM-DD. This is mainly because `Intl.DateTimeFormat` formats date based on the locale. So I can revert back to this if it works.

hellais:
> But the format is MM/DD/YYYY instead of the current YYYY-MM-DD.

Hum that won't work, since it's a non-standard locale specific way of representing dates.

We should find a way to display the time and date using the ISO 8601 (the current format is that) date and time format which is an international and unambiguous way of displaying dates.

Doing very little research on this I came across this: https://stackoverflow.com/questions/25050034/get-iso-8601-using-intl-datetimeformat, which claims that we should be choosing a specific locale which just happens to be close to ISO 8601, which does not seem ideal.

Is there a specific reason for us to use `Intl.DateTimeFormat` for formatting the date instead of using a custom date and time formatter passed through `moment` or similar tooling?

SaptakS:
@hellais right, `Intl.DateTimeFormat`  doesn't have a way to show in iso8601 format. @sarathms what should I do here? I understand that moment won't take care of internationalisation like Intl. Should I revert back to using moment in this case?

sarathms:
> Is there a specific reason for us to use Intl.DateTimeFormat for formatting the date instead of using a custom date and time formatter passed through moment or similar tooling?

@hellais In general, I wanted to use a consistent date format across the app that would work seamlessly when we start supporting more languages. So I suggested we use the `formatDate` API from `react-intl` (parsed as UTC by moment) instead of the `moment.format()`. But I guess we can make an exception for the search results.

@SaptakS Lets just use `moment.utc()...` for this. To make room for the `[UTC]` maybe you can increase the width of its `Box` and reduce that of `test_name` (each by 1 unit). It shouldn't make much of a difference to the layout.

SaptakS:
@sarathms @hellais 
Made the necessary changes. This is how it looks right now:

image

PR #393, opened 2020-02-06
Add support to show Tor test measurements

### LINKED ISSUE #333: Present findings in Explorer for Tor
Depends on: https://github.com/ooni/api/issues/118

sarathms:
Fixes #333 

**Preview**

image

@hellais please help resolving these:
* Do we deem it as an OK (`green7`) measurement only if {for all x/y, x==y} is true for bridges and directory authorities? And an anomaly in all other cases? Or do we just show the information like in the screenshot (blue hero)?
* I have relaxed the width for the name column unlike in `probe-desktop`. In any case, it doesn't work well on mobile. How about using a wrapping row for mobile layouts? The `react-table` hook we use use has [a plugin to render the table using Flexbox](https://github.com/tannerlinsley/react-table/blob/master/docs/api/useFlexLayout.md) (non `<table>` rendering) What do you think?


hellais:
> Do we deem it as an OK (green7) measurement only if {for all x/y, x==y} is true for bridges and directory authorities? And an anomaly in all other cases? Or do we just show the information like in the screenshot (blue hero)?

Yes I think we should do green if ALL of the obfs4 proxy bridges are reachable AND ALL the directory authorities are also reachable.

If some of them are not reachable, we should go for anomaly orange.

@NullHypotheis @holantonela what do you think? Does this seem reasonable?

> I have relaxed the width for the name column unlike in probe-desktop. In any case, it doesn't work well on mobile. How about using a wrapping row for mobile layouts? 

I need to research this a bit to understand if wrapping a column is an acceptable solution. My intuition though tells me that it's maybe not a good idea and we should rather be hiding or truncating some information.

I suggest we land this as it is and file an issue for how we should improve the mobile layout of the table.

holantonela:
The table needs some adjustments:
- [ ] the text should be aligned to the left (LTR should be contemplated).
- [ ] we can fix the header position so when the user scroll doesn't get lost in rows.
- [ ] in small screen devices we can fix the first column. Scrolling right will allow users to see all the table.
- [ ] zebra stripes improve readability. I highly recommend it. We could also use less hard black in the dividers.
- [ ] do we need to render the 64 characters' names? Can we strip that name? something like `abc1...5efg` might work. Users can get the entire name via JSON.

We can start with these suggestions and see how next iteration looks.

hellais:
@holantonela Thanks for your input on this. They all seem like reasonable suggestion. I will comment on a few on which I have some doubts.

Also the design of this changed a bit in the last weeks.

Here is a screenshot of the latest:
Screenshot 2020-03-05 at 15 45 21

>  in small screen devices we can fix the first column. Swiping left will allow users to see all the table.

This is for desktop only.

I think we should probably have a point after which you cannot resize the window because many things will break. The minimum size of the window is such that it should work well on any desktop screen (the default window size is 1008x646px).

hellais:
@holantonela thanks for your input. I filed a new issue for this in here: https://github.com/ooni/explorer/issues/420.

We probably also want to consolidate this view so we don't duplicate code across desktop and explorer.

Probe Desktop currently look like this: 
Screenshot 2020-03-05 at 15 45 21

PR #395, opened 2020-02-11
Show Call to Action when no data available

### LINKED ISSUE #291: Add call to action to download OONI Probe when no data is available
It was suggested by a community member that it would be useful to add a call to action on the country pages when no data is available.

In general we should also sprinkle call to actions to download OONI Probe in various other pages too.

sarathms:
Fixes #291

Currently looks like this:
image

@agrabeli You can update the copy for the key "Country.Overview.CallToAction". I just used the text from the linked issue.

@hellais Let me know what changes to make to the styling of box.


hellais:
This is great!

I think the call to action box could do with some styling improvements, though.

I made a mockup for how I think it should look like here: https://sketch.cloud/s/4WvGZ/a/kz1Z2D

components/country/OverviewCharts.js
```diff
@@ -31,24 +32,33 @@ const StyledTestGroupSelector = styled(Flex)`
...
-      </StyledCallToActionBox>
+      </Box>
+      <Flex alignItems='center' mx={4} my={4} flexDirection={['column', 'row']}>
+        <Box mr={4} mb={[3, 0]}>
+          <NLink href='https://ooni.org/install'><a>
+            <Button><FormattedMessage id='Country.Overview.NoData.Button.InstallProbe' /></Button>
+          </a></NLink>
+        </Box>
+        <Box>
+          <Button hollow><FormattedMessage id='Country.Overview.NoData.Button.OoniRunLink' /></Button>
```
hellais: As I was mentioning on the channel, I think we should table the OONI Run link button for the moment and just go for the OONI Probe Install link.

Having an OONI Run link which just points to the country list, is effectively the same as having people tap on the websites card, so we should think more about what it is we want to support doing on this page.

sarathms: Dropped button for OONI Run

Issue #403, opened 2020-02-18
Use the isAnomaly and Confirmed from the fastpath or batch pipeline


hellais:
We need to align the colors and anomaly status of the listMeasurements to the measurement pages.

The logic for calculating the screens needs to be minimal and explorer should not make any assumptions about the results, but just render what is presented to it.

If the anomaly == true, but blocking is false, it will say "Unknown" as blocking reason.

This is related to #360 .

sarathms:
This is not what was addressed in #401 right?

hellais:
Correct, this was not addressed in #401, though some additional robustness, similar to that, should be done as part of this work.

Basically we should use the `isAnomaly` and `isConfirmed` value from the pipeline to pick the color and anomaly flag (how we were doing it before), but make the code more robust such that when `blocking: false`, but `isAnomaly == true`, the pages don't break.

We can discuss this 1-1 later if you need more information.

hellais:
This has been implemented for the web_connectivity test. Closing

PR #404, opened 2020-02-21
Fix inconsistent msmt pages

sarathms:
This PR contains proposed fixes to the issues discussed recently.
* Restructuring the code in `WebConnectivity.js` measurement page component
* Injecting the message from `scores.msg` for WhatsApp measurement pages

### DIFF
--- a/components/measurement/nettests/WebConnectivity.js
+++ b/components/measurement/nettests/WebConnectivity.js
@@ -251,11 +251,7 @@ QueryContainer.propTypes = {
 
 const WebConnectivityDetails = ({
   isConfirmed,
-  // NOTE: We stop using the `isAnomaly` flag coming from `/api/v1/measurements`
-  // because it causes this bug https://github.com/ooni/explorer/issues/374
-  // where measurements are marked anomaly despite `accessible: true, blocking: false`
-  //
-  // isAnomaly,
+  isAnomaly,
   isFailure,
   country,
   measurement,
@@ -324,9 +320,7 @@ const WebConnectivityDetails = ({
         }}
       />
     )
-  } else if (/* isAnomaly */ blocking !== false) {
-    // We ignore the isAnomaly value from the API to work around
-    // https://github.com/ooni/explorer/issues/374
+  } else if (isAnomaly) {
     status = 'anomaly'
     summaryText = (
       <FormattedMessage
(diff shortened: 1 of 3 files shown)


components/measurement/nettests/WebConnectivity.js
```diff
@@ -287,10 +283,10 @@ const WebConnectivityDetails = ({
...
   }
 
-  let status = 'default',
-    // XXX: `reason` will soon be used in Hero
-    reason = null,
-    summaryText = ''
+  let status = 'default'
+  let summaryText = ''
+
+  const reason = messages[`blockingReason.${blocking}`] && intl.formatMessage(messages[`blockingReason.${blocking}`])
```
hellais: Yes this is good!

components/measurement/nettests/WebConnectivity.js
```diff
@@ -338,13 +330,12 @@ const WebConnectivityDetails = ({
           WebsiteURL: input,
           network: probe_asn,
           country: country,
-          BlockingReason: <strong>{messages[`blockingReason.${blocking}`] && intl.formatMessage(messages[`blockingReason.${blocking}`])}</strong>
+          BlockingReason: reason && <strong>{reason}</strong>
```
hellais: Maybe we don't even have to do this `&&`, because worse case we create an empty `<strong></strong>` element.

pages/measurement.js
```diff
@@ -50,7 +50,7 @@ export default class Measurement extends React.Component {
       initialProps['isAnomaly'] = results[0].anomaly
       initialProps['isFailure'] = results[0].failure
       initialProps['isConfirmed'] = results[0].confirmed
-
+      initialProps['message'] = results[0].scores.msg || null
```
hellais: I think you should probably inject into the `measurement` the whole `scores` dictionary and we can maybe for the time being, for debugging purposes, just log it to the console.

Eventually I think we would like to show it somewhere in the measurement page itself, probably somewhere close to the heading.

The `scores` field is what you are going to be using in order to extract the WhatsApp related that are calculated by the fastpath, see: https://github.com/ooni/pipeline/pull/297/files#diff-88c24525382330f3fc745dfea1834e78R765.

Basically you will get a data structure like this:
```
"scores": {
      "analysis": {
        "registration_server_accessible": true,
        "whatsapp_web_accessible": false,
        "whatsapp_endpoints_accessible": true,
    }
}
```

Which map to what you have in the WhatsApp test as: https://github.com/ooni/explorer/blob/master/components/measurement/nettests/WhatsApp.js#L30

components/measurement/nettests/WebConnectivity.js
sarathms: I generally try not to let unnecessary elements into the DOM. It has led layout issues in the past, although not exactly with this type of an element.

sarathms: That said, I don't claim that none have slipped past me :grin:

pages/measurement.js
sarathms: Done

hellais:
This is the PR addressing: https://github.com/ooni/explorer/issues/403

PR #405, opened 2020-02-21
Use fastpath scores to render WhatsApp msmt page

sarathms:
Instead of using the status flags in test_keys, we now use the  data from fastpath to determine if the endpoints, registration server or web access is blocked for whatsapp test.


components/measurement/nettests/WhatsApp.js
```diff
@@ -13,37 +13,25 @@ import MdPersonAdd from 'react-icons/lib/md/person-add'
...
-const WhatsAppDetails = ({ measurement, render }) => {
+const WhatsAppDetails = ({ isAnomaly, scores, measurement, render }) => {
   const testKeys = measurement.test_keys
   const tcp_connect = testKeys.tcp_connect
-  const registrationServerBlocked = testKeys.registration_server_status === 'blocked'
-  const webBlocked = testKeys.whatsapp_web_status === 'blocked'
-  const endpointsBlocked = testKeys.whatsapp_endpoints_status === 'blocked'
+  const registrationServerAccessible = scores.analysis.registration_server_accessible
+  const webAccessible = scores.analysis.whatsapp_web_accessible
+  const endpointsAccessible = scores.analysis.whatsapp_endpoints_accessible
```
hellais: This should be robust to the scores key being empty or not having any analysis key. Otherwise it will crash on any measurement which is not in the fastpath, which currently happens by default after some time.

hellais: If scores is not defined you should probably be using as fallback the testKeys.

sarathms: I have now added code to fallback to `test_keys` if `scores` is unavailable.

PR #412, opened 2020-02-26
Handle cases where test_keys.tcp_connect is missing

### LINKED ISSUE #381: TypeError: Cannot read property 'length' of undefined with telegram test
Happens with this measurement generated from miniooni: http://localhost:3100/measurement/20191004T105200Z_AS30722_V8ZfdAXLEjTfRb37osawgJRS3MZLVh91DYi1TXjVcXFQQmP9Bj

sarathms:
Fixes #381 

Addresses the issue in Telegram and WebConnectivity pages. Other places where this key is used are covered already:
[FacebookMessenger.js](https://github.com/ooni/explorer/blob/master/components/measurement/nettests/FacebookMessenger.js#L75)
[WhatsApp.js](https://github.com/ooni/explorer/blob/master/components/measurement/nettests/WhatsApp.js#L89)

Are there other keys from the spec that could go missing? Or is this only specific to `miniooni` generated measurements?

### DIFF
--- a/components/measurement/nettests/Telegram.js
+++ b/components/measurement/nettests/Telegram.js
@@ -73,8 +73,8 @@ const TelegramDetails = ({ measurement, render }) => {
                       />
                     </Box>
                   </Flex>
-                  {tcp_connect.length > 0 &&
-                    <React.Fragment>
+                  {tcp_connect && tcp_connect.length > 0 &&
+                    <div>
                       <Heading h={4}> <FormattedMessage id='Measurement.Details.Telegram.Endpoint.Status.Heading' /> </Heading>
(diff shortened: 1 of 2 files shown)


sarathms:
I wasn't able to test this fully because `/api/v1/measurement/` endpoint started failing. The fix for telegram page works. Will test the other one when measurements are available.

hellais:
> Are there other keys from the spec that could go missing? Or is this only specific to miniooni generated measurements?

In general I think you should make no assumption about the presence or absence of keys inside of the `test_keys` list.

We should also ensure that if there is an exception in rendering the test specific page we can at least show the raw JSON of the measurements and display an error as to why we were not able to rendere the test specific page.

PR #417, opened 2020-02-29
End to end testing with Cypress

### LINKED ISSUE #159: End to End Testing Strategy for OONI Explorer
Firstly, I think we should be able to come up with priority list of objectives to achieve using a test suite. Proposing a draft below to start with.

### Objectives
1. Ensure application builds (`yarn run build` runs successfully)
2. Ensure high level pages render without major errors (`pages/*.js`)
3. Breakages that occur when upgrading dependencies to major versions should be caught.

Are there more objectives we can think of? To what extent are the ones listed achievable? What can be tested from a UX perspective?

sarathms:
Fixes #159 

Objectives:
- [x] Ensure all pages render correctly
- [x] Basic search operation returns results

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,8 +4,7 @@ jobs:
   cypress-e2e:
     strategy:
       matrix:
-        # maybe drop `edge` if it fails too often
-        browser: [firefox, chrome, edge]
+        browser: [firefox, chrome]
     runs-on: ubuntu-latest
     name: E2E test on ${{ matrix.browser }}
     steps:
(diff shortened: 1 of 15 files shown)


sarathms:
Had to drop testing on edge (doesn't work on the test runner platform) and firefox because it isn't very stable yet (cypress-io/github-action#104).

So right now, at minimum it tries to render every page {home, search, country, countries, measurement} at least once. For measurement pages, it tests rendering for all the tests, an accessible one and an anomaly for each test. For WebConnectivity, I went ahead and tested if the hero is rendered with the right background color. Also added a test to ensure mere clicking of `Filter Results` button doesn't break the search page.

Screenshots are taken automatically on failures and uploaded to the workflow page itself e.g [here](https://github.com/ooni/explorer/actions/runs/50865587). I have enabled uploading screenshots even if there is no error, but that will work only if we take screenshots explicitly with `cy.screenshot()` during tests.

I will create an epic for writing more elaborate tests to build upon this.

Issue #202, opened 2019-06-20
Add more meta tags to page headers


sarathms:
Currently, search engines indexing our pages are not able to extract useful information to show search result summaries (as show in screenshot below). By adding `<meta>` tags with relevant properties like `description`, `twitter:image`, `twitter:summary`, `og:*` etc., we can improve these results. Also helps when these pages are shared on social media.

![image](https://user-images.githubusercontent.com/700829/59868158-29435c00-935e-11e9-937b-7640548a4f8b.png)

hellais:
This is done.

PR #430, opened 2020-03-20
Improved meta in country pages

kronaemmanuel:
Helps with #202 

- Changed the og:description to show a summary of the measurements about the country
- Changed the og:title to show the appropriate title with current country's name in it

Possible improvements: use Intl strings instead of plain english

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -26,8 +26,8 @@ export default class Header extends React.Component {
           <meta name="twitter:site" content="@openobservatory" />
           <meta name="twitter:creator" content="@openobservatory" />
 
-          <meta name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
-          <meta property="og:description" content={description} />
+          <meta key="og:title" name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
+          <meta key="og:description" property="og:description" content={description} />
           <meta property='og:type' content='website' />
 
         </Head>
--- a/pages/country.js
+++ b/pages/country.js
@@ -1,6 +1,5 @@
 /* global process */
 import React from 'react'
-import Head from 'next/head'
 import axios from 'axios'
 import {
   Container,
@@ -21,7 +20,7 @@ import WebsitesSection from '../components/country/Websites'
 import AppsSection from '../components/country/Apps'
 import NetworkPropertiesSection from '../components/country/NetworkProperties'
 import { CountryContextProvider } from '../components/country/CountryContext'
-import { useIntl } from 'react-intl'
+import IntlHead from '../components/country/IntlHead'
 
 const getCountryReports = (countryCode, data) => {
   const reports = data.filter((article) => (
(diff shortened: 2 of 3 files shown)


components/Header.js
```diff
@@ -26,10 +26,9 @@ export default class Header extends React.Component {
           <meta name="twitter:site" content="@openobservatory" />
           <meta name="twitter:creator" content="@openobservatory" />
 
-          <meta name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
-          <meta property="og:description" content={description} />
+          <meta key="og:title" name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
```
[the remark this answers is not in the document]
kronaemmanuel: We have a default ```og:description``` in Header.js and then a custom description in Country.js.
If we don't use ```key```, then Next will give us both descriptions. ```key``` prevents duplication and only gives the latter meta tag which in this case is the one in Country.js.

components/Header.js
```diff
@@ -26,10 +26,9 @@ export default class Header extends React.Component {
...
+          <meta  key="og:description" property="og:description" content={description} />
```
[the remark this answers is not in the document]
kronaemmanuel: Ahhhh, I'm working on a version of this that uses React Intl, so this will get solved in that. :sweat_smile:

kronaemmanuel:
Hey @sarathms, I seem to have messed up the yarn.lock file and I don't know how to resolve this conflict. I think i accidentally upgraded some packages.

Other than that: I have implemented the meta tags with Intl package.

components/Header.js
```diff
@@ -26,8 +26,8 @@ export default class Header extends React.Component {
           <meta name="twitter:site" content="@openobservatory" />
           <meta name="twitter:creator" content="@openobservatory" />
 
-          <meta name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
-          <meta property="og:description" content={description} />
+          <meta key="og:title" name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
+          <meta key="og:description" property="og:description" content={description} />
```
sarathms, line 30: I think there is inconsistency (even before this PR) in how we use `property` and `name` attributes. But I have opened another issue (#450) to check that. We can retain this as it is for now.

pages/country.js
```diff
@@ -43,6 +44,49 @@ const AnimatedFlex = styled(Flex)`
 const AnimatedHeading = styled(Heading)`
   transition: all 0.5s ease;
 `
+const IntlHead = ({
```
sarathms, line 47: Let's move this component into its own file under `components/country/`

pages/country.js
```diff
@@ -100,17 +144,14 @@ export default class Country extends React.Component {
       countryCode,
       countryName,
       overviewStats,
-      reports
+      reports,
```
sarathms, line 147: can remove this

components/Header.js
kronaemmanuel: Ok, I'll look that up.

pages/country.js
kronaemmanuel: Ok, I'll do that and send another PR.

pages/country.js
kronaemmanuel: Sure.

kronaemmanuel:
The issues you pointed out have been solved but I ran ```yarn``` and now I have a conflicting ```yarn.lock``` file. Can you take a look at why that would be the case?

kronaemmanuel:
@sarathms, I've reset the ```yarn.lock``` file. This can be merged now :smile:

PR #459, opened 2020-05-01
overview charts are rendered in country page

NidhiKJha:
This pull request fixes #458 
The overview charts in the country page are rendered .

PR #456, opened 2020-04-28
Implemented measurement page for Ndt7

kronaemmanuel:
#452 

Changes done for Ndt7 measurements
- [x] updated packet loss to be retransmit rate
- [x] removed outoforder
- [x] showing a ~ with ping to show that its an estimate
- [x] removed timeout
- [x] Changed the method for getting server name

The Ndt4/5 measurements still look the same.

I tried to have as little replication of code as i could without making the code a mess to read.

PR #449, opened 2020-04-05
Case sensitivity of country codes

NidhiKJha:
This pull request fixes #164 
It handles case senstivity in URLs where /country/al can be treated as /country/AL and updates the URL shown in the browser.

PR #429, opened 2020-03-20
JSON View Data overflows in search small screen

NidhiKJha:
This Pull request solves #428 

Changes:
* Solves the overflow of JSON Data in mobile screen

PR #460, opened 2020-05-05
Release 2.0.8

sarathms:
### Added
- Support for ndt7 measurements (#456)
- Meta tags for country pages (#430)
- End-to-end tests with Cypress (#417)

### Fixed
- Missing overview charts in country page (#459)
- Country codes in URL for country pages are not case sensitive anymore (#449)
- Raw measurement data content overflow in smaller screens (#429)
- Handle cases where `test_keys.tcp_connect` is missing (#412)
- Handle cases where `test_keys.requests` is missing (#412)

### Changed
- `nextjs` dependecy upgraded to 9.3.2 (#448)
- Upgrade base docker image to `node:12` as required by `next@9.x.x`
- Use `blue9` color from theme for footer background (#411)
- Updated copy

### Removed
- Dropped usage of locale-data from `react-intl`; deprecated (in v3)

### Security
- Pinned version for `https-proxy-agent` to `^2.2.3` as per github security advisory


sarathms:
Merging this although the test fails for the search page. I have verified that the page works as expected. The test tails because the API response didn't return in time.

PR #469, opened 2020-06-13
Add links to test descriptions on measurement pages

### LINKED ISSUE #282: Link to test descriptions in measurement pages
To enable users to learn more about the test (how it works, and its limitations) that each measurement is based on, it would be great if the test names in individual measurement pages linked to their respective test descriptions (available each here: https://ooni.io/nettest/). 

For example, it would be great if "Web Connectivity Test" (illustrated in the measurement page below) linked to this: https://ooni.io/nettest/web-connectivity/

blocked-site-ir

sarathms:
Fixes #282

### DIFF
--- a/components/measurement/DetailsHeader.js
+++ b/components/measurement/DetailsHeader.js
@@ -1,11 +1,12 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import prettyMs from 'pretty-ms'
-import { Flex, Box } from 'ooni-components'
+import { Flex, Box, Link } from 'ooni-components'
 // FIXME: Include 'fontWeight' to ooni-components/atoms/Text
 // Using Text from rebass directly for now
 import { Text } from 'rebass'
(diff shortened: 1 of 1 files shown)


hellais:
Looks good to me!

components/measurement/DetailsHeader.js
```diff
@@ -43,7 +44,13 @@ const DetailsHeader = ({testName, runtime, notice}) => {
           />
         </Box>
         <Box ml={2}>
-          <Text fontSize={20}>{metadata.name}</Text>
+          <Link href={metadata.info} target='_blank'>
```
hellais, line 47: I now realised that the behaviour is that of opening it in a new window.

I don't think we should be doing that. If the user wants to open it in a new window they can alt-click on the link.

hellais: Opened this PR for it: https://github.com/ooni/explorer/pull/471

PR #470, opened 2020-06-16
Improving Metatags

kronaemmanuel:
- Added MeasurementHead
- Renamed IntlHead to CountryHead
- `content` passed to SummaryText has been refined to now be an object so its easy to distinguish between already formatted strings and strings that need to passed to `intl.formatMessage()`


sarathms:
@kronaemmanuel Can you share screenshots of how this is rendered in the `<head>` section of the page? When I tried, `og:description` showed up as 
```
On {date}, {WebsiteURL} presented signs of {BlockingReason} on {network} in {country}.Check out OONI Explorer for more details and measurements
```

kronaemmanuel:
Yes, the problem seems to be in Anomaly Pages of Web Connectivity test. Other tests are fine. I'll redo the Web Connectivity Tests and submit a PR along with screenshots.

kronaemmanuel:
I made a new headMetatag component which manages all the meta tag info instead of relying on SummaryText because that was causing issues where stuff in SummaryText didn't make sense to be in headMetatag. I believe that this way is more 'standardized' even if it has some amount of duplication of code.

sarathms:
Closing this PR because it seems to have included commits from branches already merged to master. @kronaemmanuel will be opening a fresh PR.

PR #475, opened 2020-07-17
Website Centric Stats Page

### LINKED ISSUE #474: Implement website centric mockups for country listing & website graph
First iteration based on the mockups

sarathms:
Fixes #474 

#### TODO

- [x] Improve the logic for calculating percentages for each country
- [x] Add a content loader for the Global data
- [x] Add a title or lead text to the country specific page
- ~[ ] Use `swr`'s timeout features to update page when query times out~ (not that important for this issue)

Screenshot from 2020-07-17 06-37-15
Screenshot from 2020-07-17 06-34-54


sarathms:
@hellais Can you help me understand how to calculate the percentages for each country? [This](https://github.com/ooni/explorer/pull/475/files#diff-ca0ad5724df4616bee0566372cb9e1feR35-R58) is how I am doing it right now, which I realized is not what you may have intended. Nor does it seem like a sound way to compare the countries.

We also need some content/text before the country specific charts (2nd screenshot). To help the rebuild the context for you, here's what data is available on this page:
* Country Code / Name
* Website
* Date Range

And the of course the data we receive from the aggregation API.

components/aggregation/website/CountrySummary.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  if (confirmed_count > ok_count) {
+    outcome = {
+      color: colorConfirmed,
+      subtext: 'confirmed blocked',
+      percent: Number(confirmed_count / measurement_count * 100).toFixed(0),
+      icon: <IoCloseCircled size={ICON_SIZE} />
+    }
+  }
```
hellais, line 58: The percentage for a country should be:

Total of anomalous measurement (anomaly_count + confirmed_count) / total_measurement_count

We should display in the list first countries which have a confirmed count that is non-zero, then it should be sorted by the the percentages.

It would be useful to have input from @FedericoCeratto on this as well.

hellais:
@sarathms 
> Can you help me understand how to calculate the percentages for each country?

I replied in an inline comment:
https://github.com/ooni/explorer/pull/475/files#r480950127

I am also not very convinced by this approach, but I don't have much time to think about a better way. Input from @FedericoCeratto would also help, but we are a bit swamped at the moment.

Let's address the low hanging fruit first (implement the algorythm I suggested) and document as next steps to improve it based on future iterations.

> We also need some content/text before the country specific charts (2nd screenshot).

Yes I agree. What is your question?

bassosimone (commented):
We should have a chat later today about this code. I think I need some guidance to think about it in a more structured way, as my JavaScript knowledge is slowly improving. 📚

sarathms:
Rebased this branch over latest master. Merging it based on discussion with @hellais. We will collect feedback on how to make this page more useful and continue developing it based on what we find.

PR #476, opened 2020-07-27
Sharing buttons on both Country and Measurement pages

kronaemmanuel:
Things to be discussed:

- Design changes?
- We can include a default text with the link on twitter. Should we include something like: "Check out this measurement on OONI Explorer: {link goes here}" ?

### DIFF
--- a/components/SocialButtons.js
+++ b/components/SocialButtons.js
@@ -1,0 +1,21 @@
+import React from 'react'
+import { Link, Box } from 'ooni-components'
+import { Text } from 'rebass'
+import { TiSocialFacebook, TiSocialTwitter } from 'react-icons/lib/ti'
+
+export default function SocialButtons({ url, fontSize, color }){
+  return(
+    <>
+      <Box px={1}>
+        <Link color={color} target="_blank" href={`https://www.facebook.com/sharer/sharer.php?u=https://explorer.ooni.org/${url}`}>
+            <TiSocialFacebook size={fontSize}/>
+        </Link>
+      </Box>
+      <Box px={1}>
+        <Link color={color} target="_blank" href={`https://twitter.com/intent/tweet?url=https://explorer.ooni.org/${url}`}>
+            <TiSocialTwitter size={fontSize}/>
+        </Link>
+      </Box>
+    </>
+  )
+}
--- a/components/SocialButtons.js
+++ b/components/SocialButtons.js
@@ -1,21 +1,26 @@
 import React from 'react'
-import { Link, Box } from 'ooni-components'
+import PropTypes from 'prop-types'
+import { Link, Flex } from 'ooni-components'
 import { Text } from 'rebass'
-import { TiSocialFacebook, TiSocialTwitter } from 'react-icons/lib/ti'
+import { MdShare } from 'react-icons/lib/md'
 
-export default function SocialButtons({ url, fontSize, color }){
+function SocialButtons({ url }){
+  const text = 'Data from OONI Explorer';
   return(
-    <>
-      <Box px={1}>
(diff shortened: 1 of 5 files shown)


kronaemmanuel:
Country Page sharing buttons:
image

Measurement Page sharing buttons:
image

hellais:
Great job on this!

I have some suggestions about the design.

The issues I see with the current approach are:
1. The icons are a bit hidden and they don't stand out much
2. It's not clear what is going to happen when you click on them, because of 1., but also because the context might also reference social media itself (ex. we are talking about the blocking of the website twitter.com or the blocking of the facebook app)
3. I am not a fan of us promoting the logos of facebook and twitter
4. We should be absolutely sure users understand we will not track them and this is not the facebook/twitter evil tracking button, but our own tracker free option!

To this end I propose the following designs which I think address all the points above:

StreamingIconCountry2

StreamingIconMeasurement

The share icon should be inside of the material design icon set and it's called `share-variant`.

Let me know if you have any questions about the implementation.

hellais:
TODO, following call with @sarathms:
 
- Ensure that no external assets are loaded from facebook on first load.
- Write copy for the boilerplate text in the share on facebook or twitter action (cc @agrabeli )
- How does this behave on mobile? Is the design responsive?
- What happens on mobile when you click on the links, does it trigger the app or what type of workflow occurs?

components/SocialButtons.js
```diff
@@ -0,0 +1,21 @@
...
+export default function SocialButtons({ url, fontSize, color }){
+  return(
+    <>
+      <Box px={1}>
+        <Link color={color} target="_blank" href={`https://www.facebook.com/sharer/sharer.php?u=https://explorer.ooni.org/${url}`}>
+            <TiSocialFacebook size={fontSize}/>
+        </Link>
+      </Box>
+      <Box px={1}>
+        <Link color={color} target="_blank" href={`https://twitter.com/intent/tweet?url=https://explorer.ooni.org/${url}`}>
```
hellais, line 15: TODO add support for boilerplate text inside of the target.

hellais:
Here is the mockup for how the country page screenshot should look like:

Country page screenshot

components/SocialButtons.js
sarathms: Please note that this is possible only for Twitter. Facebook used to allow this earlier, but now they changed the behavior to pull the contents from the "og:*" meta tags.
 https://developers.facebook.com/support/bugs/357750474364812/?comment_id=286474311528313

https://developer.twitter.com/en/docs/twitter-for-websites/tweet-button/guides/web-intent

kronaemmanuel:
Completed the changes. Here are the new screenshots:
Country Desktop:
image

Country Mobile:
image
image

Measurement Desktop:
image

Measurement Mobile (this whole page could use some redesign on mobile btw): 
image

components/SocialButtons.js
kronaemmanuel: I've added text for twitter which says: 'Data from OONI'

hellais (approved):
This is great work @kronaemmanuel ! Thanks for putting it together.

I checked the screenshots and it looks good to me!

@sarathms can you deploy this to our testing instance of OONI Explorer so I can do some functional testing as well, before we deploy it in production?

sarathms:
Deployed. Note that it had to be merged with the branch using test API server.

sarathms:
Also the social media previews may not _look_ as good without the other metatag changes. Especially on Facebook.

sarathms:
@hellais Did you see the preview on [explorer-test](https://explorer-test.ooni.io/measurement/20190724T000050Z_AS3239_rLs5v5ZWS7oIjJhiV7LP8llLlgzwuLPz4CENYYXEHkEwN1Mj4k?input=https://gpgtools.org/)? Shall we merge this and consider tweaks for another PR?

sarathms (approved):
Ready to merge :whale:

PR #478, opened 2020-08-07
Custom metadata for measurement pages

kronaemmanuel:
- Measurement pages metadata being handled with HeadMetadata
- Country pages metadata being handled with CountryHead(instead of IntlHead)

### DIFF
--- a/cypress/integration/measurement.e2e.js
+++ b/cypress/integration/measurement.e2e.js
@@ -8,6 +8,30 @@ describe('Measurement Page Tests', () => {
   const ooniBlue = 'rgb(5, 136, 203)'
 
   describe('Web Connectivity tests', () => {
+    it('renders a valid accessible og:description', () => {
+      cy.visit('/measurement/20200807T220702Z_AS9009_VDIirQFXzvVZXGTDXrRKAd7oQB3CpnKGISOZLs7kQFV6RJNR7n')
+      cy.get('head meta[property="og:description"]')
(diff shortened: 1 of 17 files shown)


cypress/integration/measurement.e2e.js
```diff
@@ -3,6 +3,30 @@
...
+      cy.get('head meta[property="og:description"]')
+        .should('have.attr', 'content', 'On August 07, 2020, 10:35 PM UTC, http://www.google.com/search?q=lesbian was accessible in United States, explore more details and other measurements on OONI Explorer.')
+    })
+
+    // it('renders a valid website down og:description', () => {
+    //   cy.visit('')
+    //   cy.get('head meta[property="og:description"]')
+    //     .should('have.attr', 'content', '')
+    // })
+
```
hellais, line 29: Move all the above blocks into the existing test cases.

sarathms (changes requested):
This is excellent work @kronaemmanuel :clap: :+1:  I have left some comments. Most of it is about indentation/formatting. Everything else seems good.

Another suggestion is to split the end-to-end tests for measurement pages into separate files, for each nettest. Even better if they can be organized into a directory. Not sure if Cypress allows that.

sarathms (approved):
This is good to go. I have reviewed and rebased the code to work with the latest master branch. One step away from social media awesomeness. Good work @kronaemmanuel

PR #479, opened 2020-08-24
Replace api.ooni.io address with new one

hellais:


sarathms:
The tests fail because the API endpoint isn't reachable at the moment.

hellais:
@sarathms is this still relevant? Should we close it?

sarathms:
We went ahead with this change in another PR and have now also reverted to `api.ooni.io`. Closing this.

PR #483, opened 2020-08-31
Remove tested since date on country pages

### LINKED ISSUE #482: Remove tested since in explorer country page
@FedericoCeratto mentioned that we will remove the             `"first_bucket_date": first_bucket_date,` key from the private API call: https://api.ooni.io/api/_/country_overview?probe_cc=IT, because it's expensive to compute.

We should therefore drop this from the UI of OONI Explorer.

sarathms:
Fixes #482 

Removes from the date from the overview text and the meta description tag.

@agrabeli Can you please take a look at the copy and confirm if that change is acceptable? If not, can you please update the value for the key **Country.Overview.SummaryTextTemplate** in the [copy spreadsheet](https://docs.google.com/spreadsheets/d/1QAeDVF5wYj7bnTE6rJ0XX3C8sBF0o6A0TxeLw9fpTvM/edit#gid=0)?

As of now, we have been duplicating the same text in the meta description that comes up when we share the country pages on social media. If you feel, it needs a different copy, feel free to edit the text against **Country.Meta.Description** in the sheet.

image

Meta description (e.g: twitter)

image


bassosimone (approved):
🐳

PR #484, opened 2020-09-03
Dont use next/link for footer links which are external

sarathms:
Eliminates annoying errors in the browser console in dev environment.

Earlier `next/link` would work fine even if the target `href` is an external URL. Now it doesn't.
Related to: https://github.com/vercel/next.js/blob/master/errors/invalid-href-passed.md


sarathms:
Going ahead and merging this because it doesn't have any impact on UI.

PR #486, opened 2020-09-04
Refactored measurement page using refactored API

### LINKED ISSUE #489: Make sure explorer uses the refactored API
Tracks work in https://github.com/ooni/explorer/pull/486

sarathms:
Fixes #489

Uses the new `/api/v1/measurement_meta` API to fetch measurement data.
- [X] ~Revert 8d0f1c7 before merging~ We decided to continue to use the `ams-pg.ooni.org` hostname for now.

### DIFF
--- a/pages/measurement.js
+++ b/pages/measurement.js
@@ -38,12 +38,14 @@ export async function getServerSideProps({ query }) {
     params['input'] = query.input
   }
 
-  let msmtResult = await client.get('/api/v1/measurement_meta', {
+  let response = await client.get('/api/v1/measurement_meta', {
     params
   })
 
-  if (msmtResult?.data) {
-    initialProps = Object.assign({}, msmtResult.data)
+  // If response `data` is an empty object, the measurement was
+  // probably not found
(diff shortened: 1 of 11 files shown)


bassosimone:
@sarathms let's take at look at this together later today? 🙏

pages/measurement.js
```diff
@@ -106,42 +127,49 @@ export default class Measurement extends React.Component {
...
-    )
-  }
+      )}
+    </Layout>
+  )
 }
 
 Measurement.propTypes = {
-  measurement: PropTypes.object.isRequired,
+  measurement: PropTypes.object,
```
bassosimone, line 167: @sarathms please make sure we list all the props here

bassosimone (commented):
This code looks good to me. We should probably take another look at it later today, to exchange further ideas, but it's good work and definitely on the right track. Thank you!

bassosimone (approved):
I think we should merge now and without reverting anything. Because we only deploy releases, our master branch represents a development snapshot towards the goal we have in mind for the next release. Therefore, I believe it's in our best interest to bring this good set of changes to master such that we can immediately start using them along with others.

🐳

PR #488, opened 2020-09-10
Show search results indicators

### LINKED ISSUE #425: Restore anomaly flag indicators for IM tests in the search results if available
Currently the fastpath is marking as anomaly measurements coming from IM and circumvention tests.

Since those heuristics were not available in the past we had disabled it in OONI Explore.

We should now re-enable though only in the even of it being present (i.e. don't show it on batch pipeline measurements)

sarathms:
Fixes #425 

Enables indicators on IM tests. To filter out measurements coming from batch pipeline, the condition used is to check if `scores` key has object with at least one property. This goes with the assumption that fastpath measurements have non-zero properties in `scores` object.

Should we be extending this to the circumvention test too?

### DIFF
--- a/components/search/ResultsList.js
+++ b/components/search/ResultsList.js
@@ -4,7 +4,7 @@ import url from 'url'
 import moment from 'moment'
 import NLink from 'next/link'
 import styled from 'styled-components'
-import { defineMessages, FormattedMessage, useIntl } from 'react-intl'
+import { defineMessages, useIntl } from 'react-intl'
 import {
   Flex, Box,
   Link,
@@ -38,7 +38,16 @@ const ResultTagHollow = styled(StyledResultTag)`
 `
 
 const testsWithStates = [
-  'web_connectivity'
+  'web_connectivity',
+  'telegram',
+  'whatsapp',
+  'facebook_messenger'
+]
+
+const imTests = [
+  'telegram',
+  'whatsapp',
+  'facebook_messenger'
 ]
 
 const messages = defineMessages({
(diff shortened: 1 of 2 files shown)


hellais:
> Should we be extending this to the circumvention test too?

Yes I think this would be a great idea, but I suggest doing it incrementally and iteratively. As in let's land then and then do the other after this is merged.

hellais:
I tested it and it seems ok. Yet when I navigate to the search page from the homepage via CSR (instead of SSR) I get the following error:

Screenshot 2020-11-03 at 15 38 23

hellais:
The console reveals the following:
```
Access to XMLHttpRequest at 'http://api.ooni.io/api/_/test_names' from origin 'http://localhost:3100' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
_next/static/development/pages/_app.js?ts=1604414298225:6744 Object
_next/static/development/pages/_app.js?ts=1604414298225:6746 Object
api.ooni.io/api/_/test_names:1 Failed to load resource: net::ERR_FAILED
search:1 Access to XMLHttpRequest at 'http://api.ooni.io/api/_/countries' from origin 'http://localhost:3100' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
api.ooni.io/api/_/countries:1 Failed to load resource: net::ERR_FAILED
breadcrumbs.js:59 [HMR] bundle rebuilding
style.css:1 Failed to load resource: the server responded with a status of 404 (Not Found)
```

sarathms:
Why is it connecting to `api.ooni.io`?

hellais:
I fixed it by using `https` instead of `http`

hellais:
Now requires some rebase, @sarathms can you look into that?

sarathms:
Yes

hellais (changes requested):
Needs rebasing or conflict resolution

sarathms:
@hellais rebased with latest master. no changes to the commits in this branch.

hellais (approved):
LGTM

hellais:
I documented future work inside of: https://github.com/ooni/explorer/issues/520

PR #491, opened 2020-09-18
Handle missing test keys

### LINKED ISSUE #416: Ensure pages render minimal content when `test_keys` are missing
Currently pages crash when `test_keys` dictionary doesn't contain keys being used to render the page. One such bug in Web Connectivity measurement page was reported #414 and fixed recently.

A `try...catch` based approach to this problem is generally handled in React by using the [Error Boundaries](https://reactjs.org/docs/error-boundaries.html). But this doesn't work well in SSR environments like Next.js.

sarathms:
Fixes #416 

Makes sure all measurement pages render without crashing when any of the keys  in `test_keys`are not found in the measurement.

Also includes a development toolbar (not added to any pages) that allows us to control the values inside different `test_keys`
image

### DIFF
--- a/components/measurement/nettests/HTTPHeaderFieldManipulation.js
+++ b/components/measurement/nettests/HTTPHeaderFieldManipulation.js
@@ -9,7 +9,7 @@ export const HttpHeaderFieldManipulationDetails = ({ measurement, render }) => {
   const testKeys = measurement.test_keys
   let isAnomaly = false
   let isFailed = true
-  const tampering = testKeys.tampering
+  const tampering = testKeys?.tampering || {}
   Object.keys(tampering).forEach((key) => {
     if (tampering[key] === true) {
       isAnomaly = true
@@ -18,7 +18,7 @@ export const HttpHeaderFieldManipulationDetails = ({ measurement, render }) => {
       isFailed = false
     }
   })
-  const headerDiff = testKeys.tampering.header_name_diff
+  // const headerDiff = testKeys.tampering.header_name_diff
 
   return (
     render({
@@ -32,7 +32,7 @@ export const HttpHeaderFieldManipulationDetails = ({ measurement, render }) => {
       details: (
         <div>
           {/*<Text>isAnomaly: {isAnomaly.toString()}</Text>
-          <Text>isFailed: {isFailed.toString()}</Text>
+            <Text>isFailed: {isFailed.toString()}</Text>
           <Text>headerDiff: {headerDiff.toString()}</Text>*/}
         </div>
       )
--- a/components/measurement/nettests/HTTPHeaderFieldManipulation.js
+++ b/components/measurement/nettests/HTTPHeaderFieldManipulation.js
@@ -18,7 +18,6 @@ export const HttpHeaderFieldManipulationDetails = ({ measurement, render }) => {
       isFailed = false
     }
   })
-  // const headerDiff = testKeys.tampering.header_name_diff
 
   return (
     render({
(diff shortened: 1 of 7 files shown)


components/measurement/nettests/HTTPHeaderFieldManipulation.js
```diff
@@ -18,7 +18,7 @@ export const HttpHeaderFieldManipulationDetails = ({ measurement, render }) => {
       isFailed = false
     }
   })
-  const headerDiff = testKeys.tampering.header_name_diff
+  // const headerDiff = testKeys.tampering.header_name_diff
```
bassosimone, line 21: I suggest to remove code we don't need rather than commenting it out. We have git history and we are doing a great job w/ commits, so it shan't be a problem in the future to go back and ask our past self a hand.

components/measurement/nettests/HTTPHeaderFieldManipulation.js
```diff
@@ -32,7 +32,7 @@ export const HttpHeaderFieldManipulationDetails = ({ measurement, render }) => {
       details: (
         <div>
           {/*<Text>isAnomaly: {isAnomaly.toString()}</Text>
-          <Text>isFailed: {isFailed.toString()}</Text>
+            <Text>isFailed: {isFailed.toString()}</Text>
```
bassosimone, line 35: Same as above. I'd like to encourage you to not keep commented out code. Commented out code rots twice as fast as comments describing implementations, so I'd be mostly pleased if we can avoid it.

components/measurement/nettests/Ndt.js
```diff
@@ -55,27 +56,44 @@ const NdtDetails = ({ measurement, render }) => {
...
+        mss={mss}
+        packetLoss={packetLoss}
+        outOfOrder={outOfOrder}
+        timeouts={timeouts}
+      />
+    )
+  } catch (e) {
+    console.error(`Error in parsing test_keys for ${measurement.test_name}`)
+    console.error(e)
+    // Leaves performanceDetails `null` and thus isn't rendered
```
bassosimone, line 96: Thanks for this comment!

components/measurement/useTestKeyController.js
```diff
@@ -0,0 +1,71 @@
+import React, { useState, useCallback } from 'react'
+import { Flex } from 'ooni-components'
+
+// Custom hook to render a toolbar to enable/disable parts of `testKeys`
+// Usage: Add this to the MeasurementContainer
```
bassosimone, line 5: Again, I'd like to praise this correct usage of comments to explain the purpose of commented out code

components/measurement/useTestKeyController.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const TestKeyController = () => {
+    const onChange = useCallback((event) => {
+      const {name, checked} = event.target
+
+      if (name === 'all') {
+        setTestKeys(checked ? testKeysInitial : {})
+        setKeysMap(setValues(testKeysInitial, checked))
+        return
+      } else {
```
bassosimone, line 40: I'd like to encourage you to avoid `if () { ... return } else { ... }`. Why don't we follow instead the Okkam razor's principle where we don't include an else branch when we don't need it?

components/measurement/useTestKeyController.js
bassosimone: I'd perhaps just rephrase the comment to say "This file ..." such that it's clear what is the purpose of this very useful piece of code, which otherwise may not be fully obvious at a first glance.

bassosimone (approved):
Thank you, I really like what I see. Please, consider applying the recommended changes, which mainly amount to tiny changes to slightly improve reliability. 🐳

components/measurement/useTestKeyController.js
sarathms: Done. Another way to do this would be to use the `NODE_ENV === 'development'` flag to make it available in dev mode. But this isn't that frequently used either.

components/measurement/useTestKeyController.js
sarathms: Thanks. You are right. I have now changed it to a simple exclusive `if ... then` structure. Earlier I wrote some common steps at the end, which eventually changed.

sarathms:
I have addressed the comments. I made a [small change](https://github.com/ooni/explorer/pull/491/commits/7bfb746e4c6fdb9fe44cf096ced107a7f3ab9795#diff-7cde3a620c68a1903572470a0c0c8525) that makes it easy to follow the [usage described here](https://github.com/ooni/explorer/pull/491/commits/55b6a8efc35f3ebaf2dd22fc59b0a9b4b1985a13#diff-97b0c9b852cc7bf9fc615ad99c41e79e). Since it is a trivial one, I am going ahead and merging this.

PR #505, opened 2020-10-20
Handle AAAA answers in DNS queries

### LINKED ISSUE #500: DNS queries with AAAA answers causes web_connectivity measurement pages to crash
[example measurement](https://explorer.ooni.org/measurement/20201002T173034Z_AS0_X5fQaoPjGfg73APaJSMCGjAd5gSWSG07SHEZtXHnDb4yvpvU1u?input=http%3A%2F%2Fwww.iea.org%2F)

Problem: We only look for DNS response types `A` and `CNAME`, for rest of the types, we pass in an object to a component (`DnsAnswerCell`) that can only render strings, components or elements.

Possible solutions:
1) Make `DnsAnswerCell` handle an object in its `props.children` to render
2) Have logic to handle all possible `answer_type` values. Refer https://github.com/ooni/spec to know what to expect
3) Fallback values to use when the `answer` is corrupt

sarathms:
Fixes #500 

Now we show ipv4 for `A` type, `ipv6` for `AAAA` and `hostname` for `CNAME` and ignore everything else.

### DIFF
--- a/components/measurement/nettests/WebConnectivity.js
+++ b/components/measurement/nettests/WebConnectivity.js
@@ -60,6 +60,11 @@ const StatusInfo = ({ url, message}) => (
   </Flex>
 )
 
+StatusInfo.propTypes = {
+  url: PropTypes.string,
+  message: PropTypes.string
+}
+
 // From https://css-tricks.com/snippets/css/make-pre-text-wrap/
 const WrappedPre = styled(Pre)`
   white-space: pre-wrap;       /* css-3 */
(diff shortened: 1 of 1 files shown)


components/measurement/nettests/WebConnectivity.js
```diff
@@ -287,12 +304,6 @@ const WebConnectivityDetails = ({
     timeZone: 'UTC',
     timeZoneName: 'short'
   })
-  const reasons = {
-    'http-diff': 'HTTP-diff',
-    'http-failure': 'HTTP-failure',
-    'dns': 'DNS',
-    'tcp_ip': 'TCP'
-  }
```
bassosimone, line 295: I assume this was unused correct?

bassosimone (approved):
🐳 

asked a question just to be sure I understand

components/measurement/nettests/WebConnectivity.js
sarathms: Yes

PR #512, opened 2020-10-27
Set default since date in search to 30 days from until date

### LINKED ISSUE #507: Set the default since date in search to a month back from current date
This is to align with the change in the default date range on API when fetching measurements, which is now set to 30days from current date. We will now set the default value of since field to `${today}-30`.
The page URL will also reflect the same just like what we do for the until field.

sarathms:
Fixes #507 

* Going directly to `/search` sets the since date to `${today} - 30`
* If there is already an `until` date in the page URL, since date is set to `${until} - 30`

image

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -144,6 +144,11 @@ class Search extends React.Component {
       query.until = until
     }
 
+    const since = moment(query.until).utc().subtract(30, 'day').format('YYYY-MM-DD')
+    if (!query.since) {
+      query.since = since
+    }
+
     [testNamesR, countriesR] = await Promise.all([
       client.get('/api/_/test_names'),
       client.get('/api/_/countries')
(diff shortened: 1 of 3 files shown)


pages/search.js
```diff
@@ -144,6 +144,11 @@ class Search extends React.Component {
       query.until = until
     }
 
+    const since = moment(query.until).utc().subtract(30, 'day').format('YYYY-MM-DD')
+    if (!query.since) {
+      query.since = since
+    }
```
bassosimone, line 150:
```suggestion
    if (!query.since) {
      query.since = moment(query.until).utc().subtract(30, 'day').format('YYYY-MM-DD')
    }
```

bassosimone (approved):
🐳 The fix looks good to me. I have tested the new implementation at the testing instance by checking for the following test cases:

1. no filters

1. only filter on until and verify since is set correctly

1. filter on both and verify they're unchanged

1. filter only on since and verify until is set to today

PR #522, opened 2020-11-11
feat: add basic page for riseupvpn

bassosimone:
Part of https://github.com/ooni/backend/issues/473

Next step: computing if there are errors.


components/measurement/nettests/RiseupVPN.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types'
+
+const RiseupVPNDetails = ({ measurement, render }) => {
+  // TODO(bassosimone,cyberta): possibly refine the algorithm
```
bassosimone, line 4: @cyberta: does this look good to you as a first stab at scoring the measurement?

sarathms (approved):
This looks good. :whale: 
The only thing missing is the metadata for the new test (name, group, url) in `components/test-info.js`. I will add it because it needs some work on the copy as well.

sarathms:
The logic to use a default blank measurement page with just the raw data was broken. I added a fix. With this, any measurement from tests that explorer doesn't recognize yet will still be shown like this:

image

cyBerta:
Nice! I didn't follow this development the last days and I'm very happy to see this progress!

PR #523, opened 2020-11-17
Switch to using the hostname api.ooni.io

### LINKED ISSUE #508: chore: point explorer to api.ooni.io
Now that the flip is complete, we should modify the codebase to use api.ooni.io.

sarathms:
Fixes #508


sarathms:
Github action fails because of an unrelated problem documented in #524 . The tests themselves pass according to the logs.

PR #529, opened 2020-12-04
Add link to measurements in overview text on country page

### LINKED ISSUE #338: Link to measurements on OONI Explorer Country pages
The OONI Explorer country pages provide aggregate information (which is great!), but they no longer share the bulk of all individual measurements collected from that country. The old OONI Explorer used to do that, which is why I used to reference OONI Explorer country pages in research reports. Now, if I reference an OONI Explorer country page, it doesn't share the measurements collected from that country in an easy-to-find way (only some links are included that point to measurements for specific tested resources). 

I therefore propose that we add a button (or something that makes it obvious and easy to find) at the start of each OONI Explorer country page which links to the Search page for that country, enabling users to easily/quickly gain access to all measurements collected from that country.

For example, the Egypt page (https://explorer.ooni.org/country/EG) would have a link (that is easy to find) to: https://explorer.ooni.org/search?until=2019-10-26&probe_cc=EG

sarathms:
Fixes #338 

The link points to say `/search?probe_cc=IN` which would then rewritten to `/search?probe_cc=IN&until=2020-12-05&since=2020-11-05` when the page gets loaded to reflect that we are only searching within the last 30 days.

image


hellais (approved):
LGTM

PR #530, opened 2020-12-04
Bump version of github actions to v2

### LINKED ISSUE #524: Upgrade github actions dependencies to v2
E2E tests show up as failed even though the test themselves pass because of an error thrown in one of the core github actions that runs at the beginning. This is related to [a recent security bugfix](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) at Github. This can be addressed by upgrading the 3 dependent github actions like `actions/checkout@v1` to `v2`.

sarathms:
Fixes #524

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


sarathms:
Merging this directly because it is a trivial change to the github actions workflow config.

PR #531, opened 2020-12-09
Show monthly coverage stats skipping last month data

sarathms:
Before:
image

After:
image

### DIFF
--- a/components/landing/Stats.js
+++ b/components/landing/Stats.js
@@ -1,3 +1,4 @@
+/* global process */
 import React from 'react'
 import {
   VictoryChart,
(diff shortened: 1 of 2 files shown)


components/landing/Stats.js
```diff
@@ -29,45 +30,37 @@ const getMaxima = (data) => {
...
+    measurementsByMonth.pop()
 
     // API responses are ordered by date, with most recent month at the end
     const lastMonth = {
-      countryCount: countryCoverage[countryCoverage.length - 1].value,
-      networkCount: networkCoverage[networkCoverage.length - 1].value,
-      measurementCount: measurementsByMonth[measurementsByMonth.length - 1].value
+      countryCount: countryCoverage[countryCoverage.length - 2].value,
+      networkCount: networkCoverage[networkCoverage.length - 2].value,
+      measurementCount: measurementsByMonth[measurementsByMonth.length - 2].value
```
hellais, line 63: Maybe it's more robust to ignore any datum which is in the future doing a filter based on what is the current month.

PR #543, opened 2021-01-28
Handle missing array values in test_keys

sarathms:
This fixes the problem in pages like https://explorer.ooni.org/measurement/20210128T090057Z_webconnectivity_IT_30722_n1_WOS6jGJBP1UMopM9?input=http%3A%2F%2Fvoice.yahoo.jajah.com%2F where test_keys has null values in `tcp_connect`, `queries.[].answers` ﻿
We explicitly check that the value exists and is an `Array`


sarathms:
I have also added a test for the reported page.

bassosimone (approved):
:whale:

PR #544, opened 2021-02-03
Bump up the font-size of the footer

hellais:
Before
Screenshot 2021-02-03 at 16 54 46
After
Screenshot 2021-02-03 at 16 54 51

The footer was previously way too small, this PR bumps it up to a more reasonable font size.


sarathms (approved):
:whale:

PR #546, opened 2021-02-04
Improve the labels of the mat

hellais:
This PR rotates the date labels by 45 degrees so they become more legible and moves the y axis label outside of the number area to improve legibility.


sarathms (approved):
LGTM! :+1:

PR #549, opened 2021-02-19
Validate raw measurement

sarathms:
This merges the `raw_measurement` received from API with an object with acceptable defaults that lets the rest of the page render conditionally.

This is written only for web_connectivity now. If the pattern works well, I can extend it or do the validation in a higher level component for all measurements.

Test deployment: https://explorer-test.ooni.io/measurement/20201001T132904Z_AS3737_Eg5Snk8fPivqLkexBjmlmczIdVLCjatSDFXptj60xrTgav6eOj?input=https://www.fidh.org/

Note: Started using `??` operator a lot. See [Nullish Coalescing Operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). This could replace a lot of other places where there is problematic code like
```javascript
falseCouldBeValid && <Comp uses={falseCouldBeValid} />
```


sarathms:
Sample page with fallback text for missing information.

image

sarathms:
@hellais I left the `Download JSON` button available because the URL to access it is still valid. Although in most cases, it won't show any data. It can also be a quick way to check what the API server says. e.g. in this case
image

components/measurement/DetailsBox.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react'
```
sarathms, line 1: Better prop-type definitions and refactored to functional components.

components/measurement/nettests/WebConnectivity.js
```diff
@@ -271,18 +272,45 @@ QueryContainer.propTypes = {
   query: PropTypes.object
 }
 
+const validateMeasurement = (measurement) => {
```
sarathms, line 275: This validation function can either be evolved into a generic one to run before deciding to render a specific component from `measurement/nettests/*` or similar local methods across all other measurements. Right now it makes sure the component works with an object that has all the expected keys, even if they are absent in API responses.

sarathms: Maybe I should add this in the file itself.

components/measurement/nettests/WebConnectivity.js
```diff
@@ -294,7 +322,7 @@ const WebConnectivityDetails = ({
       dns_experiment_failure,
       control_failure
     }
-  } = measurement
+  } = validateMeasurement(measurement ?? {})
```
sarathms, line 325: This handles the case where `measurement` itself is a non-object.

components/measurement/nettests/WebConnectivity.js
```diff
@@ -505,19 +532,23 @@ const WebConnectivityDetails = ({
...
                 title={<FormattedMessage id='Measurement.Details.Websites.DNSQueries.Heading' />}
                 content={
-                  <React.Fragment>
-                    <Flex flexWrap='wrap' mb={2}>
-                      <Box mr={1}>
-                        <strong><FormattedMessage id='Measurement.Details.Websites.DNSQueries.Label.Resolver' />:</strong>
-                      </Box>
-                      <Box>
-                        {client_resolver || '(unknown)'}
+                  Array.isArray(queries) ? (
```
sarathms, line 535: A more uniform pattern to show the `No Data` label in the boxes when keys like `queries`, `tcp_connect` are invalid or missing.

cypress/integration/measurement.e2e.js
```diff
@@ -100,7 +100,7 @@ describe('Measurement Page Tests', () => {
         .should('have.attr', 'content', 'On August 7, 2020, Whatsapp was reachable in United States, explore more details and other measurements on OONI Explorer.')
     })
 
-    it.only('renders an unreachable og:description', () => {
```
sarathms, line 103: This should ideally be caught in CI by including `yarn lint` in the workflow. There are quite a few unaddressed lint errors (mostly missing prop-types) that would render it useless if enabled.

pages/measurement.js
```diff
@@ -65,15 +65,10 @@ export async function getServerSideProps({ query }) {
         }
       }
 
-      if (typeof initialProps['raw_measurement'] === 'string') {
```
sarathms, line 68: This slightly reverses another recent attempt to deal with missing keys in API responses. Instead of failing early when `raw_measurement` is null, we let it pass so that we can at least render the metadata-based UI and show whatever can be shown.

hellais (approved):
This pull request looks good to merge! I think it significantly improves the robustness of the measurement pages.
I really like the use of the `??` operator which I did not know of the existence.

Issue #547, opened 2021-02-15
Set correct title tags in measurement pages


hellais:
The title of measurement pages is currently simply set to `<title>OONI Explorer</title>`, while we should instead construct it based on the description and content of the measurement.

hellais:
This is done.

PR #550, opened 2021-02-19
Add meta description to measurement pages

hellais:
This implements:
* https://github.com/ooni/explorer/issues/547
and part of https://github.com/ooni/explorer/issues/202


components/Header.js
```diff
@@ -1,37 +1,41 @@
...
-          <meta property='og:type' content='website' />
-
-        </Head>
-      </header>
-    )
-  }
+const description = 'OONI Explorer is an open data resource on Internet censorship around the world consisting of millions of measurements on network inteference.'
+
+const Header = () => {
+  const canonical = 'https://explorer.ooni.org' + useRouter().pathname;
```
sarathms, line 9: extra `;`

components/Header.js
```diff
@@ -1,37 +1,41 @@
...
+        <meta property="og:url" content={canonical} />
+
+        <meta name="twitter:card" content="summary_large_image" />
+        <meta name="twitter:title" content="OONI Explorer" />
+        <meta name="twitter:site" content="@openobservatory" />
+        <meta name="twitter:creator" content="@openobservatory" />
+
+        <meta key="og:title" name='og:title' property='og:title' content='OONI Explorer - Open Data on Internet Censorship Worldwide' />
+        <meta key="og:description" property="og:description" content={description} />
+        <meta property='og:type' content='website' />
```
sarathms, line 35: We should also add an `og:image` tag. According to [OGP spec](https://ogp.me/#metadata) it is one of the required properties.

components/measurement/HeadMetadata.js
```diff
@@ -15,10 +15,10 @@ const HeadMetadata = ({
   const intl = useIntl()
   let description = ''
 
+  const formattedDate = moment.utc(date).format('LL')
```
sarathms, line 18: We should reduce using `momentjs` and use the native Intl APIs via `react-intl` more. This can be replaced
```suggestion
  const formattedDate = intl.formatDate(date, { day: 'numeric', month: 'long', year: 'numeric'})
```

components/measurement/HeadMetadata.js
```diff
@@ -32,10 +32,17 @@ const HeadMetadata = ({
 
   return (
     <Head>
+      <title>
```
sarathms, line 35: Why can't we include date in the title also?

components/measurement/nettests/Dash.js
```diff
@@ -63,7 +63,8 @@ const getOptimalQualityForBitrate = (testKeys) => {
 const messages = defineMessages({
   dash: {
     id: 'Measurement.Metadata.Dash',
-    defaultMessage: 'See results of Dash Test on {date} in {country} and other measurements on OONI Explorer.'
+
```
sarathms, line 66: This shows up as 
image

components/measurement/nettests/Tor.js
```diff
@@ -235,7 +235,7 @@ const TorDetails = ({
   const messages = defineMessages({
     tor: {
       id: 'Measurement.Metadata.Tor',
-      defaultMessage: 'See results of Tor Test on {date} in {country} and other measurements on OONI Explorer.'
+      defaultMessage: 'Tor censorship test result in {country}'
```
sarathms, line 238: Why not use the `isAnomaly` flag?

sarathms (approved):
I think this is good to go. I added some comments. Most of those I will address. Couple of things to discuss:

1. About including an `og:image` property. This can be just the OONI logo in `1:1`. But maybe we can include this when merging the PR with screenshots as `og:image` that Krona wrote.
2. Is there a reason why descriptions for `Tor` was left generic and not deterministic like `Psiphon`?

components/measurement/HeadMetadata.js
hellais: Because otherwise it ends up being too long. The [recommended length for the title tag is less than 64 characters](https://www.w3.org/Provider/Style/TITLE.html)

components/measurement/nettests/Tor.js
hellais: I just copied what already existed and I didn't want to make any changes. If we can easily refactor it to include different text for anomaly that would be great.

components/measurement/nettests/Dash.js
hellais: Good call. I suspect similar grammar breakage will occur on the NDT test result as well. The cleanest thing to do would be to separate the `description` from the `title` and handle the differently in every page, but it's going to be quite a bit more work.

Is there some way to extract the actual video speed inside of the test result and display that in the description?

Issue #541, opened 2021-01-14
Inconsistent date/time formats in measurment page summary texts


sarathms:
On `web_connectivity` pages we say 
```
On March 02, 2020, 9:17 PM UTC, http://prachatai.com was accessible when tested on AS15751 in Ireland.
```
On IM test pages we show only the date, with the full date-time format in a tooltip. ([example measurement](https://explorer.ooni.org/measurement/20200807T220116Z_AS9009_JaRyaMzjdCq5JA89mE1nl8VCZltDv4vAaNhUqi193SFCcwR8av))
```
On August 7, 2020, WhatsApp was reachable on AS9009 in United States. 
```

Note: The same pattern was adopted into the `<meta>` tags as well.

sarathms:
This was fixed in #568

PR #558, opened 2021-03-17
Fixed inconsitent date/time formats in measurement page summary texts…

Eddges:
As stated in issue #541, it only showed the date on IM test pages
```
On March 17, 2021, WhatsApp was reachable on AS7922 in United States
```
However in the web_connectivity pages it said
```
On March 02, 2020, 9:17 PM UTC, http://prachatai.com was accessible when tested on AS15751 in Ireland.
```

It will now show Date _and_ Time in both pages, adding to consistency.

### DIFF
--- a/components/measurement/SummaryText.js
+++ b/components/measurement/SummaryText.js
@@ -16,7 +16,7 @@ const SummaryText = ({
 }) => {
   const intl = useIntl()
   const metadata = getTestMetadata(testName)
-  const formattedDate = moment(date).format('LL')
+  // const formattedDate = moment(date).format('LL')
   const formattedDateTime = intl.formatDate(moment.utc(date).toDate(), {
     year: 'numeric',
     month: 'long',
@@ -37,7 +37,7 @@ const SummaryText = ({
           testName: `[${metadata.name}](${metadata.info})`,
           network: network,
           country: country,
-          date: `<abbr title='${formattedDateTime}'>${formattedDate}</abbr>`
+          date: `<abbr title='${formattedDateTime}'>${formattedDateTime}</abbr>`
         }}
       />
   } else {


sarathms (changes requested):
:clap: Thanks @Eddges for working on this. This change does make it consistent. Coincidentally, this just happens to be the way we wanted to make it consistent (not the other one where we hide the time on all pages).

However, I would request you to make one more change before we can merge this. Now that we are rendering the full time stamp on the page, we can drop the tooltip feature with `<abbr>` which isn't necessary anymore.

components/measurement/SummaryText.js
```diff
@@ -16,7 +16,7 @@ const SummaryText = ({
 }) => {
   const intl = useIntl()
   const metadata = getTestMetadata(testName)
-  const formattedDate = moment(date).format('LL')
+  // const formattedDate = moment(date).format('LL')
```
sarathms, line 19: remove the commented line

components/measurement/SummaryText.js
```diff
@@ -16,7 +16,7 @@ const SummaryText = ({
...
   const formattedDateTime = intl.formatDate(moment.utc(date).toDate(), {
```
sarathms, line 20: Can you try and do this without having to use the `moment.utc` call? In general we would like to move away from [moment](https://momentjs.com), so we are trying to refactor the code that uses the package. You can find more information about using [alternatives here](https://momentjs.com/docs/#/-project-status/).

If this isn't a straight forward change, don't worry about it. Just reply here that it is better to defer it to a future PR or something specific you have to say about it.

Eddges: Hey, thanks for reviewing! 
I have done the changes. I am looking into removing momentJS dependency altogether with something lightweight. Will it be okay if I open a future PR after removing this dependency from the project completely?

Also, I noticed I have committed changes directly to the master branch in this PR. I have made sure to make branches for the other PRs.

Issue #487, opened 2020-09-08
Use transient props in styled-components to get rid of DOM warnings


sarathms:
Components extended using `styled-components` pass through the props we use for conditional styling (e.g `isOpen`) down to the underlying DOM triggering warnings in the browser console. This has been addressed in `styled-components@v5.1.x` by providing [transient props](https://styled-components.com/docs/api#transient-props) which are consumed by SC and won't be passed down.

As a prerequisite, **we have to upgrade `styled-components`**, which was not always very smooth.

Here's one place where this happens: [CollapseTrigger.js](https://github.com/ooni/explorer/blob/master/components/CollapseTrigger.js#L9)
There are probably one or two other places where this happens.

DevRishiJain:
I am a beginner can you guide me about this isseu.

I want to work on it

PR #560, opened 2021-03-19
FIX: DOM Warnings By Using Transient Props

msaaddev:
In this PR, I have addressed issue #487. For this, I have done the following:

- [x] Updated `styled-components` to v5.1.0.
- [x] Replaced simple props with transient props where a component is extended using styled-components and have conditional styling in it.

I didn't replace simple props with transient props in [Overview.js](https://github.com/ooni/explorer/blob/master/components/country/Overview.js) file because the styled component `NwInterferenceStatus` was commented out later in the file.

I opened this PR earlier but later closed it because I accidentally committed to my master branch of the `explorer`. Now I have changed my branch, made the changes, and opened it again. :)

@sarathms kindly take a look

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,88 +3,81 @@
 
 
 "@ampproject/toolbox-core@^2.6.0", "@ampproject/toolbox-core@^2.7.1":
-  version "2.7.1"
-  resolved "https://registry.yarnpkg.com/@ampproject/toolbox-core/-/toolbox-core-2.7.1.tgz#d433a333c0dbb0bb6db7e69edd490d5694e06fc3"
-  integrity sha512-MWGmCLyBOouXTy1Vc30Jw7NkshJ5XkPlcXhhRc9Gw3dDAZJ8rUS69SIQ6cFMt2owCQnw7irMNlvZQTqdyx61rA==
+  "integrity" "sha512-MWGmCLyBOouXTy1Vc30Jw7NkshJ5XkPlcXhhRc9Gw3dDAZJ8rUS69SIQ6cFMt2owCQnw7irMNlvZQTqdyx61rA=="
+  "resolved" "https://registry.npmjs.org/@ampproject/toolbox-core/-/toolbox-core-2.7.1.tgz"
+  "version" "2.7.1"
   dependencies:
-    cross-fetch "3.0.6"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -3,88 +3,81 @@
 
 
 "@ampproject/toolbox-core@^2.6.0", "@ampproject/toolbox-core@^2.7.1":
-  version "2.7.1"
-  resolved "https://registry.yarnpkg.com/@ampproject/toolbox-core/-/toolbox-core-2.7.1.tgz#d433a333c0dbb0bb6db7e69edd490d5694e06fc3"
-  integrity sha512-MWGmCLyBOouXTy1Vc30Jw7NkshJ5XkPlcXhhRc9Gw3dDAZJ8rUS69SIQ6cFMt2owCQnw7irMNlvZQTqdyx61rA==
+  "integrity" "sha512-MWGmCLyBOouXTy1Vc30Jw7NkshJ5XkPlcXhhRc9Gw3dDAZJ8rUS69SIQ6cFMt2owCQnw7irMNlvZQTqdyx61rA=="
+  "resolved" "https://registry.npmjs.org/@ampproject/toolbox-core/-/toolbox-core-2.7.1.tgz"
```
sarathms, line 7: This is probably because your setup is different. Check if it is because of using `npm` instead of `yarn` or some specific configuration in `yarn`.

I would suggest that you restore this file to the version in `master` and run `yarn` again to get an updated file. If this isn't working, then it is better to not include any changes to `yarn.lock`. I have to update the file in `master` anyway after merging with other changes.

sarathms (changes requested):
@msaaddev Thank you for opening this PR. I'm glad `styled-components@v5` was a non-breaking upgrade. I left a couple of comments with some suggestions. We can merge this as soon as they are addressed. Awesome work!

sarathms:
Good that you followed the practice of using branches to open PRs. :+1:

yarn.lock
msaaddev: Yes, I used `npm` instead of `yarn`. I will try to quickly fix it now. 🙌

msaaddev: @sarathms I have reverted the file to the master version. It should be good to go!

Issue #468, opened 2020-05-11
Present "valid domain" message when domain is entered with backslash


agrabeli:
A community member shared the following feedback:

1) If you enter the domain name incorrectly in the search field like this "http://www.gayvideo.ru/", OONI Explorer gives you a proper warning like the one shown in the scan below:

![explorer-1](https://user-images.githubusercontent.com/8290574/81582994-02947680-93b1-11ea-8810-c587f8dc4573.png)

2) If you enter the domain name up properly like this "www.gayvideo.ru",  OONI Explorer shows you the measurements:

![explorer-2](https://user-images.githubusercontent.com/8290574/81583102-248df900-93b1-11ea-82a3-3d4f4894f99a.png)

3) There is a third scenario when you can enter the domain name properly but forget to remove the backslash at the end of the URL like this"www.gayvideo.ru/" Then OONI Explorer does not give you any warnings, outputs right away "No results found" message and suggest to play with the filters to get better results. The filters in the last two examples stay the same - the only element that varies is the backslash "/"

![explorer-3](https://user-images.githubusercontent.com/8290574/81583168-3bcce680-93b1-11ea-8d5b-99263fab6c71.png)

Maybe you can tweak the settings of the Explorer to give researchers warning when they enter the domain name with a backslash.

majakomel:
Closing this ticket since the issue seems to be resolved. If we need any more adjustments to the validation let's create a new ticket.

Issue #457, opened 2020-04-30
Domain field validation passes for invalid domains


hellais:
For example if you filter by `google.com/` it passes the validation, yet produces no results.

Here is an example:
![Screenshot 2020-04-30 at 18 59 18](https://user-images.githubusercontent.com/424620/80738094-c5a8c400-8b14-11ea-9590-222f411b4f16.png)

majakomel:
I tried to reproduce, but this seems to be resolved now.

hellais:
Yes you are right, I think we can consider this issue solved.

PR #562, opened 2021-03-19
FIX: Valid Domain Issue when used backslash

msaaddev:
In this PR, I have addressed issues #468 and #457. 

## Problem

OONI Explorer is searching and giving no results for a domain that has backslash `/` at the end. 

search with backslash

When searched without backslash `/`, it was giving the desired results.

Search without backslash

## Solution

I have improved the regex so that now it is giving the appropriate error message.

Solution

### DIFF
--- a/components/search/FilterSidebar.js
+++ b/components/search/FilterSidebar.js
@@ -144,7 +144,7 @@ class FilterSidebar extends React.Component {
       case 'domainFilter':
         var domainValue = e.target.value
         // eslint-disable-next-line no-useless-escape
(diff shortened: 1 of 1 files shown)


msaaddev:
@sarathms one of the tests is failing. I am not exactly sure why. Can you please take a look?

sarathms:
> one of the tests is failing.

It was the API not returning enough data. I have rerun the job. Lets see if it goes well this time. Thanks for checking.

msaaddev:
@sarathms does this PR require any more work?

majakomel:
Closing this since linked tickets are closed already (issue seems resolved)

PR #564, opened 2021-03-19
[ENHANCEMENT] Date and time format inconsistency in the search results

Eddges:
This is in response to the issue #515 

This PR:
- [x] Fixes the inconsistency in the date and time formats between search result elements and the measurements page

As of now, the measurements page presents the date and time in this format: 

Screenshot from 2021-03-19 18 11 12

And the Search results have a different format for showing the same:

Screenshot from 2021-03-19 18 10 38

This PR aims to bring a consistency in the format of the displayed date and time:

Screenshot from 2021-03-19 18 10 56

### DIFF
--- a/components/search/ResultsList.js
+++ b/components/search/ResultsList.js
@@ -330,7 +330,17 @@ const ResultItem = ({
                   <ASNBox asn={probe_asn} />
                 </Box>
                 <Box width={5/16}>
-                  {moment.utc(measurement_start_time).format('YYYY-MM-DD HH:mm [UTC]')}
+                  {
+                    useIntl().formatDate(measurement_start_time, {
+                      year: 'numeric',
+                      month: 'long',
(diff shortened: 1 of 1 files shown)


hellais:
Thanks for opening this PR!

The correct fix for this issue should be that of using the `measurement_start_time` key inside of the measurement page (similarly to the search page) instead of the `test_start_time` key.

sarathms (changes requested):
@Eddges The problem starts in `pages/measurement.js` where we are using `test_start_time` from the API response. I think doing a simple search-and-replace on all instances of `test_start_time` with `measurement_start_time` should just work. Make sure you change the prop names also.

The change you made can also be retained if you can restore the output format in the search page as before `YYYY-MM-DD HH:MM UTC` without using `moment`. We want to use `intl.formatDate` everywhere.

Issue #298, opened 2019-09-12
Update Search Tool image on landing page


agrabeli:
Currently the Search Tool image on the landing page is outdated as it's based on the UX of the old Search page. 

![search](https://user-images.githubusercontent.com/8290574/64766117-c967e600-d545-11e9-8ccc-d6d8fac4c961.png)

Replace this image with a new one that reflects the new UX of the Search page.

PR #566, opened 2021-03-19
[ENHANCEMENT] Updated Search section's image on the homepage

Eddges:
With reference to the issue #298

The current homepage search section's image is still based on the older UI

Screenshot from 2021-03-19 22 25 09

So I made an image based on the new UI 

Screenshot from 2021-03-19 22 25 21

The link to the Figma file is 
https://www.figma.com/file/4kZHrb9VxBr2rqX2hE0FHw/OONI-Search?node-id=1%3A2

Feedbacks are welcome! I'll do the changes based on your feedbacks :)

Please review @agrabeli  @sarathms  @hellais


Eddges:
@sarathms I'm not sure why the test has failed here. I checked with e2e test on my local machine and it passed without any problem (all specs passed). Can you please take a look? The only changes to the code are
1. Copying a new image to the assets directory
2. Changing the static image link in the index page

sarathms:
> @sarathms I'm not sure why the test has failed here. I checked with e2e test on my local machine and it passed without any problem (all specs passed). Can you please take a look?

Looks like this was also because of irregular response from API. I ran the tests again and it has passed now.

sarathms (changes requested):
Thanks for working on this @Eddges. The original issue is from a long time ago. We would have never been able to get to it by ourselves :laughing: 

It looks good to me. A few minor suggestions:
1. Have a test like `Web Connectivity` selected in the screenshot.
2. The other two seem to have some amount of shadow around the screenshots. It would look better to have something similar.

I would request @hellais also to take a look and suggest improvements. A full-page screenshot can be found in the artifacts section of the [last github action](https://github.com/ooni/explorer/actions/runs/668659632) on this PR.

@Eddges: After @hellais responds, you can replace the existing `search.png` with the new one instead of adding a new file.

Eddges:
Okay, sure!

hellais:
Yes this looks good. I agree with @sarathms about adding a subtle shadow to align it to the other screenshots and replacing the file instead of adding an additional `searchUpdated.png` file.

sarathms:
@Eddges Would you like give this another shot?

Eddges:
@sarathms yes indeed, right after my exams (till 5th May). I have some work remaining in other PRs too.

Eddges:
@sarathms as per your feedback on this thread and the design channel, I have made necessary changes and have replaced the old image with the new one. Kindly check

PR #575, opened 2021-04-15
Add support for Signal measurements

### LINKED ISSUE #574: Display whether Signal measuremements are OK or present anomalies
We recently released a new OONI Probe test for Signal, and Signal measurements are openly published on OONI Explorer: https://explorer.ooni.org/search?since=2021-03-15&test_name=signal&until=2021-04-16

However, the measurements don't currently display the result of the Signal test (i.e. if it was accessible or presented an anomaly). This means that users currently need to read the raw data (under `test_keys`) to get the test result. 

We therefore need to add support for displaying the value (test result) of Signal measurements.

sarathms:
Fixes #574 

image

image


hellais (approved):
LGTM

agrabeli:
Thank you @sarathms !

PR #587, opened 2021-05-14
Move environment variable definitions to .env files

### LINKED ISSUE #492: Handle environment variables better with nextjs@9.4.x
Next.js v9.4+ uses `.env.*` files to load environment variables which should greatly simplify the way we currently consume them, mostly to pick the API server address.
Ref: https://nextjs.org/blog/next-9-4#new-environment-variables-support

sarathms:
Fixes #492 

Uses [environmental variables](https://nextjs.org/docs/basic-features/environment-variables) feature of Next.js

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+NEXT_PUBLIC_MEASUREMENTS_URL=https://api.ooni.io
+PORT=3100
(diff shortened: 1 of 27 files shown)


hellais (approved):
LGTM

PR #594, opened 2021-05-21
Test legacy pages

### LINKED ISSUE #554: End to end testing of old YAML data in explorer
This is about writing integration tests which perform a search query for results in the range of 2012-2014, try to access several measurements and ensure that the pages do not break.

sarathms:
Fixes #554 

Searches for measurements between `2014-12-01` and `2015-03-31` and checks if the measurement pages render properly and if basic information is shown on those pages.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,41 +1,28 @@
 {
-    "env": {
-        "browser": true,
-        "es6": true
+  "env": {
+    "browser": true,
+    "es6": true
+  },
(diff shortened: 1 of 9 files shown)


hellais:
This looks like a bug in the API. In theory it should return some data for the YAML based measurements. @FedericoCeratto do you know what is going on with this?

FedericoCeratto:
Due to https://github.com/ooni/backend/issues/416

PR #622, opened 2021-10-22
MAT Table View

### LINKED ISSUE #608: mat: Sort the labels of the multi-axis charts alphabetically
Some labels should be sorted alphabetically, specifically category_code, probe_cc and possibly probe_asn. We don't want to do the sorting of the labels if it's input as it would significantly slow down render time.

sarathms:
Fixes #608 

- Use clickhouse URL for aggregation API calls
- Fixed height 1-axis chart; dynamic xAxis label
- Basic Table view with virtualized rows
- Add filtering and selecting rows to the table view
- Fix layout issues on mat page
- static chart switch; filtering done right
- Better debug window
- Move charts to be a child of table


.env
```diff
@@ -1,2 +1 @@
-NEXT_PUBLIC_MEASUREMENTS_URL=https://api.ooni.io
```
hellais, line 1: Why is this no longer set inside of the default production env?

components/aggregation/Debug.js
```diff
@@ -17,9 +17,23 @@ Bold.propTypes = {
   children: PropTypes.any,
 }
 
-export const Debug = ({ query, children, ...rest }) => {
-  const { query: queryCtx, apiResponse, others, preReshapeTimeRef, reshapeTimeRef, renderTimeRef} = useDebugContext()
+const getCircularReplacer = () => {
+  const seen = new WeakSet()
```
hellais, line 21: Should we use a polyfill for this? According to caniuse this feature is missing in ~6% of web users browsers: https://caniuse.com/?search=Weakset.

hellais: Also this function seems to be duplicated over here: https://github.com/ooni/explorer/blob/ux/mat-table-view/pages/search.js#L65

components/aggregation/mat/CustomTooltip.js
```diff
@@ -7,45 +7,75 @@ import NLink from 'next/link'
...
-  const linkToMeasurements = useMemo(() => {
+  const queryObj = ['probe_cc', 'test_name', 'category_code', 'probe_asn', 'input'].reduce((q, k) => {
+    if (k in data)
+      q[k] = data[k]
+    else if (k in query)
+      q[k] = query[k]
+    return q
+  }, {})
+  if (axis_y === 'input') {
+    queryObj.input = urlToDomain(data.input)
```
hellais, line 36: You should not be converting the `input` into a `domain` for the search query. The search query works on `input` fields as well.

components/aggregation/mat/StackedBarChart.js
```diff
@@ -22,15 +39,15 @@ export const StackedBarChart = ({ data, query }) => {
       let indexBy = ''
       indexBy = query['axis_x']
       let reshapedData = Array.isArray(data.data.result) ? data.data.result.map(d => {
-        d['ok_count'] = d.measurement_count - d.confirmed_count - d.anomaly_count
+        d['ok_count'] = d.measurement_count - d.anomaly_count - d.failure_count
```
hellais, line 42: It's possible for a measurement to be anomalous, but also be marked as fail. Until https://github.com/ooni/backend/issues/486 is implemented, I think it's best to just subtract `anomaly_count` from the `measurement_count` to compute the `ok_count`.

components/aggregation/mat/StackedBarChart.js
```diff
@@ -22,15 +39,15 @@ export const StackedBarChart = ({ data, query }) => {
...
         return d
       }) : data.data.result
       return {
         data: reshapedData,
         dimensionCount: data.data.dimension_count,
         url: data.url,
         loadTime: data.loadTime,
-        cols,
+        cols: (indexBy === 'blocking_type') ? ['measurement_count'] : cols,
```
hellais, line 50: I'm not sure what this does. Can you clarify? I think we should for the time being omit logic for `blocking_type` plotting, because the data is not being formatted properly at the source and I don't think we are going to be able to properly display it.
What we really want to be doing with `blocking_type` is display it as values inside of each bar, not adding it as a new axis.

hellais: I have documented the API work to be done for this here: https://github.com/ooni/backend/issues/554

components/aggregation/mat/CustomTooltip.js
sarathms: This needs to be changed on the search page then. Since we only have UI input for `domain`, we filter out any extra URL params when constructing the API query.
e.g https://explorer.test.ooni.org/search?since=2022-01-25&until=2022-01-27&probe_cc=AF&test_name=web_connectivity&input=http%3A%2F%2Fbitcoin.org%2F&failure=false does not send the `input` param in the API call. Two changes can be made incrementally.
1. Allow `input` param to pass through to the API query
2. The above will show an empty `Domain` field but with results that are filtered. Some suitable UX solution to address this.

hellais: Ah right. Well in that case let's just go for populating this with `domain`. The code above still needs to change as it should be using `domain` instead of input in the explorer URL.

components/aggregation/Debug.js
sarathms: I believe `next.js` includes polyfill for old browsers. This is in a less frequented code path. But I will double check.

.env
sarathms: This is already included in `.env.production`

components/aggregation/mat/CustomTooltip.js
sarathms: > should be using domain instead of input in the explorer URL

components/aggregation/mat/StackedBarChart.js
sarathms: > I'm not sure what this does. Can you clarify?

sarathms: >I'm not sure what this does. Can you clarify?

For the 1-dimension chart, in case of `blocking_type` as X-axis, this replaces the stack with a simple bar showing the value in `measurement_count`. That was my interpretation of the data in the [API response](https://backend-fsn.ooni.org/api/v1/aggregation?probe_cc=AF&test_name=web_connectivity&since=2022-01-26&until=2022-01-28&axis_x=blocking_type).

Anyway, as discussed above, I will be removing `blocking_type` as an option for X-axis. So this will also be removed.

components/aggregation/Debug.js
sarathms: 1. It looks like WeakSet is polyfilled by `next.js`. I found it in the file `.next/static/chunks/polyfills.js`.
2. This code is in a dev only part (Debug.js) which will not eventually go into production. The other instance in `pages/search.js` is dead code anyway.

PR #623, opened 2021-11-10
Show failed measurements in search results

sarathms:
This was previously disabled because of a backend bug that
marked all measurements as errors. We now have more accurate flags
to show failed measurements in the search results.
﻿


sarathms:
Seems alright. Merging it.

PR #627, opened 2021-11-26
Evaluate status in the order (failure, confirmed, anomaly)

sarathms:


sarathms:
As of now this results in showing all measurements as errors on the search page with default filters.

image

hellais:
Perhaps we should exclude by default from the listing all measurements which are marked as failed. We can then have a checkbox in the filter section that optionally allows users to display also the failed measurements.

This might require some API level changes as well.

hellais:
It's actually already supported by the API via `https://api.ooni.io/api/v1/measurements?since=2021-11-01&until=2021-12-02&limit=50&failure=false`.

We should add a checkbox to the filter section which says "Hide failed measurements" and is enabled by default.

sarathms:
@hellais Added a checkbox to the search filters.
image

It is enabled by default and added to the URL if missing upon first navigation to the search page. Note that when the option is unchecked, i.e errors are *not hidden*, I am removing the `failure=??` part from the URL. In other words, there will only be a `failure=false` in the query parameters because `failure=true` shows *only* errors.

Preview: https://explorer-cej9m75mh-ooni1.vercel.app/

sarathms:
When the option is unchecked:
image

hellais (approved):
It looks like a fair amount of refactoring also happened as part of this PR. I took a look at it and it all seems like quite an improvement over the previous situation.

I did some minimal testing of the deployment and all seemed to work reasonably well.

LGTM! 🐳

PR #653, opened 2022-01-26
Dashboard for Circumvention Tools Reachability

sarathms:
Fixes ooni/ooni.org#774

Preview deployment: https://explorer.test.ooni.org/chart/circumvention


hellais (changes requested):
This is looking good!

Some things that should be fixed are:
* The link to the view measurements doesn't do filtering based on test_name.
* When I select all countries, the section for the tor snowflake test is much taller than it should be (that is to say that there is a lot of whitespace under it even though there is no chart data)
* When I select all countries and then I deselect all countries by clicking on X the country selector becomes irresponsive.
* Related to the previous point, it seems like there is a difference between choosing "Select all" and not having any country selected. While both of these selections result in all countries being visible, in one case they are rendered inside of a container, while in the other case they are rendered uncontained. We should pick one of the two and stick to that (in the first iteration we said we would keep them in the virtualized view).

hellais:
A few issues noticed in testing:
Screenshot 2022-02-28 at 15 02 30

The country selector overlaps with the chart. This only happens when you have interacted with the chart (i.e. displayed the tooltip), on first load it doesn't overlap.

Screenshot 2022-02-28 at 15 02 47
When you click on the tooltip for one of the last charts it's displayed hidden below the overflow container. Can we either make the tooltip go out of the container or if we can't do that, make it render above the click instead of below?

The top axis with the dates should be closer to the charts. 
Screenshot 2022-02-28 at 15 07 09

Also, can we have similarly to the MAT the vertical lines matching up with the Y axis?
See below for how it looks like in the MAT page:
Screenshot 2022-02-28 at 15 08 37

Ideally it would actually look like the following:
Screenshot 2022-02-28 at 15 09 59

sarathms:
@hellais I have addressed the following:
* fixed the overlapping country selector
* closing the gap between x-axis and top row chart
* enabling vertical lines (gridY)
image


The only one I postponed is the hidden tooltip, which I think can be solved by not using a scrollable container when there are less number of charts. Either by conditionally disabling virtualization itself or switching CSS for the container. I will be implementing this in a separate PR.

sarathms:
I will remove the link to the dashboard page from NavBar and merge this to master so that we can deploy some MAT improvements to production.

Issue #633, opened 2021-12-15
Search page crashes when typing into the date selector


hellais:
## Steps to Reproduce the Problem

  1. Go to the search page
  2. Select the Since date selector
  3. Select and type using the keyboard a date of choice
  4. The page crashes

## Specifications

  - Platform: Google Chrome

PR #654, opened 2022-01-27
Fix crash when manually changing the date filter in search page

hellais:
Closes: https://github.com/ooni/explorer/issues/633


sarathms (approved):
👍 Thanks for fixing this.

PR #658, opened 2022-01-28
Fix how ok_count is calculated

sarathms:
Fixes issue introduced in 0a396aa


hellais (changes requested):
The API now returns an `ok_count` value for aggregation queries. You should switch to using that.

sarathms (commented):
LGTM 👍

PR #681, opened 2022-02-28
Mat title

hellais:
This PR achieves the following:
* Add support for displaying a subtitle with metadata about the filters selected in the single axis charts
* Display a title and subtitle in the 2 axis charts
* Add legend to 2 axis charts
* Fix positioning of the OONI logo
* Load the chart with a default height of 800px


sarathms (approved):
This looks good to me. In a next iteration we should render readable versions of the labels (e.g category descriptions, test names etc).

There was one issue with the layout where the xAxis and the 2-axis charts got separated by a wide margin like this:
image

I am adding a commit that fixes this.

PR #684, opened 2022-03-04
Generate ChartHeader based on MATContext

sarathms:
* Uses a single component to show the header on the 1-axis chart, 2-axis chart and the PT dashboard charts.
* Also fixes the issue of the header title not updating when country selection is changed back to 'All Countries'.


sarathms:
Going ahead and merging this to deploy to production.

PR #685, opened 2022-03-05
Switch to using domain instead of input to filter

hellais:
We should probably eventually support both, but for the time being this is better than the current situation


sarathms (approved):
LGTM :+1:

PR #688, opened 2022-03-08
Tooltip visibility in Circumvention charts

sarathms:
This addresses the issue discussed in https://github.com/ooni/explorer/pull/653#issuecomment-1054295488

> When you click on the tooltip for one of the last charts it's displayed hidden below the overflow container. Can we either make the tooltip go out of the container or if we can't do that, make it render above the click instead of below?
> Screenshot 2022-02-28 at 15 02 47

This is solved by not using virtual lists to render the rows when the counts are low. Currently the threshold is set to `10` rows. 

image

When virtualized for longer lists, the scroll action is more natural to make such tooltips visible. However, in most case, the tooltip is anchored in a way that it is visible.


hellais:
Yes I think this is a solid approach. I tested it with more than 10 rows and the anchoring of the tooltip in the last rows works as expected (it's displayed on top instead of below).

LGTM!

PR #691, opened 2022-03-09
Circumvention charts polish

sarathms:
- [x] drop probe_cc from header
- [x] fix sorting of rows
- [x] readable labels in ChartHeader (e.g psiphon => Psiphon Test)


hellais:
This looks good to me.

The only thing I would suggest changing, is that when we disable the chart title, also disable the chart subtitle.

sarathms:
> The only thing I would suggest changing, is that when we disable the chart title, also disable the chart subtitle.

This was done in [bb9f3f4](https://github.com/ooni/explorer/pull/691/commits/bb9f3f44d495687e6136322a01371c16294ee445)

sarathms:
Couple of more changes:
* Change order of sections to: Psiphon, Tor, Tor Snowflake
* Added back the link in nav bar to access the dashboard page

Issue #689, opened 2022-03-08
Implement a better date picker


hellais:
Currently we have a date picker component that requires two interactions to select a date range.

It would be great if we switched to something that allowed the user to select a range in a single interaction and possibly also specify the range as something like "last X days".

It's important that what we choose supports the ability to also specify the start and end date by typing into the field the timestamp.

Mohamed-Ayaan358:
![image](https://user-images.githubusercontent.com/72858215/161677283-686bc71c-e690-431b-b8df-32c23cc71383.png)
Is this the relevant date time picker that we are discussing in this issue?

Issue #510, opened 2020-10-26
Replace momentjs as a dependency


sarathms:
The momentjs team has [announced](https://momentjs.com/docs/#/-project-status/) that they consider it to be a legacy project in maintenance-only mode. We may not need to replace it immediately. But we should understand the implications for us and evaluate the way forward. Some useful reads before we discuss this:
- https://inventi.studio/en/blog/why-you-shouldnt-use-moment-js
- https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore
- https://blog.logrocket.com/4-alternatives-to-moment-js-for-internationalizing-dates/

hellais:
This has been done, we might revisit it in light of the new date picker component.

PR #699, opened 2022-03-12
Replace momentjs with dayjs where possible

majakomel:
Related issue: https://github.com/ooni/explorer/issues/510

Also fixes an issue with inconsistent time display - described in: https://github.com/ooni/explorer/issues/515

Initially, I intended to go with date-fns, but unfortunately it doesn't have a nice interface to work with timezones/utc and it required working a lot with the native Date object, which defeats the purpose of a library altogether. After a bit of exploration and testing, I decided to go with day.js, which is very lightweight, just as popular, and with a very similar interface as momentjs - which makes it also easy to replace.

Unfortunately, it's not possible to get rid of moment.js completely just yet since the `Datetime` component depends on it. I will follow up with [this issue](https://github.com/ooni/explorer/issues/689), to replace it with a date-library-agnostic component and remove the use of momentjs completely in another PR.

### DIFF
--- a/components/dashboard/Form.js
+++ b/components/dashboard/Form.js
@@ -2,15 +2,16 @@ import { useEffect, useMemo, useState } from 'react'
 import { territoryNames } from 'country-util'
 import { useForm, Controller } from 'react-hook-form'
 import { Box, Flex } from 'ooni-components'
-import moment from 'moment'
 import { MultiSelect } from 'react-multi-select-component'
 import { useIntl } from 'react-intl'
+import moment from 'moment'
+import dayjs from 'services/dayjs'
 
 import { StyledLabel } from '../aggregation/mat/Form'
 import DatePicker from '../DatePicker'
 
-const tomorrow = moment.utc().add(1, 'day').format('YYYY-MM-DD')
-const lastMonthToday = moment.utc().subtract(30, 'day').format('YYYY-MM-DD')
+const tomorrow = dayjs.utc().add(1, 'day').format('YYYY-MM-DD')
+const lastMonthToday = dayjs.utc().subtract(30, 'day').format('YYYY-MM-DD')
 
 const defaultDefaultValues = {
   since: lastMonthToday,
--- a/components/landing/Stats.js
+++ b/components/landing/Stats.js
@@ -9,7 +9,7 @@ import {
   VictoryLegend
 } from 'victory'
 import axios from 'axios'
-import moment from 'moment'
+import dayjs from 'services/dayjs'
 import { Flex, Text, theme } from 'ooni-components'
 import { useIntl } from 'react-intl'
 import useSWR from 'swr'
@@ -148,9 +148,10 @@ const CoverageChart = () => {
               }
             ]}
           />
+          {/* format(parseISO(t), 'MMM\'\'yy' */}
           <VictoryAxis
             tickCount={12}
-            tickFormat={(t) => moment(t).format('MMM[\']YY')}
+            tickFormat={(t) => dayjs(t).format('MMM\'YY')}
           />
           <VictoryAxis
             dependentAxis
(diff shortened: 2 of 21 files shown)


components/measurement/SummaryText.js
```diff
@@ -16,7 +15,7 @@ const SummaryText = ({
 }) => {
   const intl = useIntl()
   const metadata = getTestMetadata(testName)
-  const formattedDateTime = intl.formatDate(moment.utc(date).toDate(), {
+  const formattedDateTime = intl.formatDate(date, {
```
hellais, line 18: I think you need to cast this to `utc` otherwise `intl.formatDate` will assume it's in the default locale of the browser resulting and then when it converts it back to UTC it will be offset by the wrong value.

As a demonstration of this see: https://explorer-git-fork-majakomel-replace-momentjs-ooni1.vercel.app/measurement/20220211T235940Z_webconnectivity_DE_8881_n1_kS8lfHqxOpVopvoi?input=https%3A%2F%2Ftheintercept.com%2F

Where it says:
```
On February 11, 2022, 10:59 PM UTC
```

yet the test_start_time value (expressed in UTC) is `"2022-02-11 23:59:39"`

components/measurement/nettests/WebConnectivity.js
```diff
@@ -334,7 +333,7 @@ const WebConnectivityDetails = ({
   } = validateMeasurement(measurement ?? {})
 
   const intl = useIntl()
-  const date = intl.formatDate(moment.utc(test_start_time).toDate(), {
+  const date = intl.formatDate(test_start_time, {
```
hellais, line 336: Same comment as above

components/dashboard/Form.js
```diff
@@ -2,15 +2,16 @@ import { useEffect, useMemo, useState } from 'react'
 import { territoryNames } from 'country-util'
 import { useForm, Controller } from 'react-hook-form'
 import { Box, Flex } from 'ooni-components'
-import moment from 'moment'
 import { MultiSelect } from 'react-multi-select-component'
 import { useIntl } from 'react-intl'
+import moment from 'moment'
+import dayjs from 'services/dayjs'
```
hellais, line 8: Is it a feature of next.js to allow absolute imports relative to the components directory tree?

hellais:

Left a comment on something that should be addressed prior to merge.

Also, while you are touching this bit of the codebase, perhaps you can also integrate a fix for this bug: https://github.com/ooni/explorer/issues/515#issuecomment-1031664899?

components/dashboard/Form.js
majakomel: It requires a bit of configuration, which is already there https://github.com/ooni/explorer/blob/master/jsconfig.json#L2-L4

components/measurement/SummaryText.js
majakomel: React Intl library was actually supposed to handle that (see https://github.com/ooni/explorer/blob/b47038a930da7a9fc7f31110999a748669ad5feb/components/measurement/SummaryText.js#L24) and I haven't noticed any problems with it locally (the link above doesn't work so couldn't check there). I ended up replacing those too with dayJS ( also because `react-intl` is currenly 2 major versions behind).

majakomel:
Thanks for the review @hellais, I fixed the issue with the timezone and inconsistent time display, can you check again if things are as expected?

components/landing/Stats.js
```diff
@@ -148,9 +148,10 @@ const CoverageChart = () => {
               }
             ]}
           />
+          {/* format(parseISO(t), 'MMM\'\'yy' */}
```
hellais, line 151: Perhaps we can remove this commented out line

hellais:
I did some testing of the branch and it works great.

I left 2 very minor comments in the PR, which once addressed we can merge it.

hellais:
Yes this looks great. Thanks for putting it together!

I am going to proceed with merging it.

Issue #690, opened 2022-03-08
Add support for displaying the probe resolver information inside of the measurement pages


hellais:
We should add support for displaying in the measurement pages the metadata about the resolver used by the probe.

These are the values stored inside of the `resolver_ip`, `resolver_network_name`, `resolver_asn` top level keys of the measurement.

It's important to take into account the fact that in some measurements some or all of these fields might be absent, since the probe isn't generating them.

PR #700, opened 2022-03-13
Show resolver info on measurement pages

majakomel:
Related ticket: https://github.com/ooni/explorer/issues/690

I added the resolver info within `CommonDetails` component since the data might be present for various probes.


components/measurement/CommonDetails.js
```diff
@@ -97,17 +100,42 @@ const CommonDetails = ({
     },
     {
       label: intl.formatMessage({ id: 'Measurement.CommonDetails.Label.Engine' }),
-      value: engine ?? intl.formatMessage({ id: 'Measurement.CommonDetails.Value.Unavailable' })
+      value: engine
```
hellais, line 103: Should probably still display unavailable when it's not defined.

hellais:

Left a small nitpick comment, but otherwise LGTM.

I noticed that the measurement page crashes when loading missing measurements from the search (this is likely caused by the fact that the testing API is used in staging deployments instead of the production). This is, though unrelated to these changes so I documented it as a separate issue: https://github.com/ooni/explorer/issues/701.

Once you address the comment I left above, we can proceed with merging it.

components/measurement/CommonDetails.js
majakomel: This is a bit of cleanup, since the variable is already initialised as `unavailable` (just like `platform`) https://github.com/ooni/explorer/blob/eff50e47a35be23167671d133ae58facc8fbf270/components/measurement/CommonDetails.js#L69-L70

hellais: Ah sorry I missed that. I think then this is good to merge!

hellais:
LGTM 🐳

public/static/lang/en.json
```diff
@@ -73,6 +73,10 @@
   "Measurement.Details.Websites.HTTP.Request.URL": "URL",
   "Measurement.Details.Websites.HTTP.Response.Body": "Response Body",
   "Measurement.Details.Websites.HTTP.Response.Headers": "Response Headers",
+  "Measurement.CommonDetails.Label.Resolver": "Resolver",
+  "Measurement.CommonDetails.Label.ResolverASN": "Resolver ASN",
+  "Measurement.CommonDetails.Label.ResolverIP": "Resolver IP",
+  "Measurement.CommonDetails.Label.ResolverNetworkName": "Resolver Network Name",
```
sarathms, line 79: I have added these keys to the spreadsheet to ensure they are not deleted the next time we sync up the `.json` files.

PR #702, opened 2022-03-14
Set different defaults to env vars in prod/dev/test envs

sarathms:
Separate defaults for env variables in different environments.
This allows new devs to use test instance of API for local development out of the box. ﻿


sarathms:
@hellais merging this directly because it is a trivial change and doesn't affect production deployments.

PR #703, opened 2022-03-15
lazy loading the country images

GermaVinsmoke:
Current case -

https://user-images.githubusercontent.com/32815509/158387128-9fb7ad14-77bd-4ff9-b065-07e58a53075d.mp4

It is loading every country image at the time of page load, I think creating too much pressure on the main javascript thread. How about showing images only after we scroll?

New lazy loading case - 

https://user-images.githubusercontent.com/32815509/158387324-f6350bbf-1293-4d51-8c62-163906edc37e.mp4

### DIFF
--- a/components/Flag.js
+++ b/components/Flag.js
@@ -48,7 +48,7 @@ export const Flag = ({countryCode, size, border}) => {
   const src = `/static/flags/1x1/${countryCode}.svg`
   return (
     <FlagContainer className='country-flag' size={size} border={border}>
-      <FlagImg src={src} size={size} />
+      <FlagImg src={src} size={size} loading='lazy' />
     </FlagContainer>
   )
 }


sarathms:
> A member of the Team first needs to authorize it

I have authorized this to find out if and how much it improves the experience on the page. Thank you @GermaVinsmoke

sarathms:
@GermaVinsmoke Can you help understand the gains from this improvement?

GermaVinsmoke:
`loading=lazy` attribute when added to the `image` tag ensures that the image loads in a lazy manner. It'll only load the images which are present in the viewport.
As the user scrolls down, it'll load up the images which come in the viewport.
So, it offloads the loading of every image present on the page in one go.

The loading time (First contentful paint) of the page reduces significantly with this technique. And also reduces the Time to Interactive factor of the website.

The technique name is "Defer offscreen images", also recommended by Google Lighthouse 😃.

https://web.dev/offscreen-images/
https://web.dev/browser-level-image-lazy-loading/

Issue #697, opened 2022-03-11
mat: add support for downloading the chart data as CSV and JSON


hellais:
We should add buttons at the bottom of the MAT chart that allows users to download the data used to plot the chart in CSV and JSON format.

This should be pretty easy as the API already supports exporting data as CSV (you just append `format=CSV`) and it's basically just a matter of adding some links pointing to those URLs.

GermaVinsmoke:
@sarathms I would like to take this.
Have few questions. 
1. What is MAT
2. Which page is having this chart
3. Is there any API documentation from where we can check all the API endpoints?

hellais:
Thanks for offering to take this on!

The MAT page is the following: https://explorer.ooni.org/experimental/mat?test_name=web_connectivity&since=2022-02-14&until=2022-03-17&axis_x=measurement_start_day&axis_y=probe_cc

The API endpoint it uses for fetching the metadata of the chart is the following: https://api.ooni.io/apidocs/#/default/get_api_v1_aggregation.

What needs to be done to implement this, is looking at the aggregation API query that is being issued for the chart and creating a link button at the bottom of the page to that API query with the `format=CSV` suffix. When a user will click on that they will get the data for the chart in CSV format.

hellais:
This has been implemented since a while.

PR #707, opened 2022-03-16
added csv and json download buttons

GermaVinsmoke:
- Created a new file in utils folder `dataExport.js` which is having functions for exporting and downloading the different types of data.
- Changed the import paths in `mat.js`
- Replaced `if-else` with **ternary** operator in `getServerSideProps`
- Created new `ReactNode (type)` functions for different types of tasks like loading, error and showing data.

Feature for #697 

### DIFF
--- a/utils/dataExport.js
+++ b/utils/dataExport.js
@@ -5,12 +5,12 @@ export const requestUrl = (query) => {
   const qs = new URLSearchParams(query).toString()
   const reqUrl = `${process.env.NEXT_PUBLIC_AGGREGATION_API || process.env.NEXT_PUBLIC_MEASUREMENTS_URL}/api/v1/aggregation?${qs}&format=CSV`
 
-  return reqUrl;
+  return reqUrl
 }
 
 export const exportToCsv = async (event, query) => {
-  event.preventDefault();
-  const url = requestUrl(query);
+  event.preventDefault()
+  const url = requestUrl(query)
 
   try {
     const response = await axios.get(url)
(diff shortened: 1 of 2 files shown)


sarathms (changes requested):
@GermaVinsmoke Thanks for picking up this important feature. The preview deployment failed because of lint errors. We use a no-semicolon style code. Do you mind fixing the errors reported by `eslint` in these two files? Thanks.

GermaVinsmoke:
@sarathms Removed the semi-colons from the file.

utils/dataExport.js
```diff
@@ -0,0 +1,22 @@
+import axios from 'axios'
+
+export const exportToCsv = async (query) => {
+  const qs = new URLSearchParams(query).toString()
+  const reqUrl = `${process.env.NEXT_PUBLIC_AGGREGATION_API || process.env.NEXT_PUBLIC_MEASUREMENTS_URL}/api/v1/aggregation?${qs}&format=CSV`
+  try {
+    const response = await axios.get(reqUrl)
+    const element = document.createElement('a')
+    element.href = 'data:text/csv;charset=utf-8,' + encodeURIComponent(response.data)
```
hellais, line 9: Since these files can become quite large I suggest we don't encode them using dataURI, but rather just provide a direct link to the CSV file. This has the added benefit that a user will be able to right click on the link and copy it to their clip board for download.

I also doesn't involve having to download the file twice in the browser.

hellais: I suggest also setting the `download="${YY-MM-DD}-ooni-data.csv` attribute so that when the user clicks on the link, it will trigger a download of it instead of opening it in the view.

GermaVinsmoke: Nevermind the above comment, I got it working.

GermaVinsmoke:
Closing this because old and this feature is already added by some other PR.

PR #709, opened 2022-03-18
MAT responsive form UI

GermaVinsmoke:
Fixes the responsive UI of the form on every device.

Previous State of the form on mobile screen -
image

New State of the form on mobile screen -
image

### DIFF
--- a/components/aggregation/mat/Form.js
+++ b/components/aggregation/mat/Form.js
@@ -52,11 +52,7 @@ export const Form = ({ onSubmit, testNames, query }) => {
 
   return (
     <form onSubmit={handleSubmit(onSubmit)}>
-      <Flex my={2} justifyContent='space-between' sx={{
-        '@media screen and (max-width: 476px)': {
-          flexDirection: 'column',
-        }
-      }}>
+      <Flex my={2} justifyContent='space-between' flexDirection={['column', 'row']}>
         <Box width={[1, 1/5]}>
           <StyledLabel>
             Country
@@ -152,11 +148,7 @@ export const Form = ({ onSubmit, testNames, query }) => {
           />
         </Box>
       </Flex>
-      <Flex my={2} justifyContent='space-between' sx={{
-        '@media screen and (max-width: 476px)': {
-          flexDirection: 'column',
-        }
-      }}>
+      <Flex my={2} justifyContent='space-between' flexDirection={['column', 'row']}>
         <Box width={[1, 0.18]}>
           <StyledLabel>
             Test Name
@@ -165,7 +157,7 @@ export const Form = ({ onSubmit, testNames, query }) => {
             name='test_name'
             control={control}
             render={({field}) => (
-              <Select {...field} sx={{ width: '100%' }}>
+              <Select {...field} width={1}>
                 {testNames.map((test, idx) => (
                   <option key={idx} value={test.id}>{test.name}</option>
                 ))}
@@ -196,7 +188,7 @@ export const Form = ({ onSubmit, testNames, query }) => {
             name='category_code'
             control={control}
             render={({field}) => (
-              <Select {...field} sx={{ width: '100%' }}>
+              <Select {...field} width={1}>
                 <option value="">ALL</option>
                 {categoryCodes.map(([code, label], idx) => (
                   <option key={idx} value={code}>{label}</option>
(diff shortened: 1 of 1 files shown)


components/aggregation/mat/Form.js
```diff
@@ -52,14 +52,18 @@ export const Form = ({ onSubmit, testNames, query }) => {
...
+        }
+      }}>
+        <Box width={[1, 1/5]}>
           <StyledLabel>
             Country
           </StyledLabel>
           <Controller
             render={({field}) => (
-              <Select {...field}>
+              <Select {...field} sx={{ width: '100%' }}>
```
sarathms, line 66: I think it is better to avoid using an object literal to pass down props for these kind of components. It causes unnecessary re-renders when the parent renders for unrelated reasons. You can instead just pass `width={1}` because `Select` is constructed with a `Flex` component.

components/aggregation/mat/Form.js
```diff
@@ -148,24 +152,28 @@ export const Form = ({ onSubmit, testNames, query }) => {
           />
         </Box>
       </Flex>
-      <Flex justifyContent='space-between'  alignItems='center'>
-        <Box>
+      <Flex my={2} justifyContent='space-between' sx={{
+        '@media screen and (max-width: 476px)': {
+          flexDirection: 'column',
```
sarathms, line 157: Same as the earlier comment about media queries.

components/aggregation/mat/Form.js
```diff
@@ -148,24 +152,28 @@ export const Form = ({ onSubmit, testNames, query }) => {
...
+        <Box width={[1, 0.18]}>
           <StyledLabel>
             Test Name
           </StyledLabel>
           <Controller
             name='test_name'
             control={control}
             render={({field}) => (
-              <Select {...field}>
+              <Select {...field} sx={{ width: '100%' }}>
```
sarathms, line 168: Same as the earlier comment about using object literal as prop. Also applies to other such instances below.

components/aggregation/mat/Form.js
```diff
@@ -148,24 +152,28 @@ export const Form = ({ onSubmit, testNames, query }) => {
...
         </Box>
       </Flex>
-      <Flex justifyContent='space-between'  alignItems='center'>
-        <Box>
+      <Flex my={2} justifyContent='space-between' sx={{
+        '@media screen and (max-width: 476px)': {
+          flexDirection: 'column',
+        }
+      }}>
+        <Box width={[1, 0.18]}>
```
sarathms, line 160: Using `0.18` doesn't have the readability as `1/6th of the row`. The standard practice is to use fractions, so that the sibling Boxes can be configured similarly (e.g `2/6`, `3/6`). This ensures that the total width doesn't exceed  `1` (or `100%`).

Also in this case, I think making it `width={1/6}` would make the element too narrow for longer options like `HTTP Header Field Manipulation` to be hardly visible in screenshots etc. I would suggest using `1/5` because there are 5 boxes in the row anyway.

sarathms (changes requested):
@GermaVinsmoke Thanks for this submission. It definitely makes the form look much better on mobile.

I have requested a couple of changes related to how we use `styled-component` in this project. The idea is to keep the notations consistent with rest of the code base. Let me know if you need any clarifications on those comments.

components/aggregation/mat/Form.js
GermaVinsmoke: The problem with `1/5` is that the width will become `20%` of all 5 elements.
That'll leave 0 space in between the elements, that's why I decided to use `18%` width which gave a nice spacing in between the elements. The `max-width` of the whole container is also `1280px` which is not giving enough space for the spacing.

How about using 9/50 instead of 0.18? 😂

sarathms: Usually I address that using padding within the elements. Although even that is can be a not-so-ideal solution at times.

sarathms:
@GermaVinsmoke Thanks for sending in the requested changes. This is a new feature and so undergoing heavy development. We decided to make some more changes to the Form component which, in a way, changes the problem definition for this PR. I will be adding a commit to address those changes and merging this along with your changes.

Looking forward to more contributions from you. Feel free to reach out if you want to discuss which one to pick up next.

sarathms:
New layout
image

On mobile:
image

sarathms (approved):

Issue #680, opened 2022-02-25
fix: remove console logs in the explorer


thara003:
There are few console logs in the production server and I think it's not a good idea to have those.

## Expected Behavior
No console logs in the production server

## Actual Behavior
Here, we can see a few console logs in the  production server 

![image](https://user-images.githubusercontent.com/79561774/155764108-b331529d-62ec-494e-94d3-c88af60489e7.png)

## Steps to Reproduce the Problem

  1. Go to [explorer search ](https://explorer.ooni.org/search?since=2022-01-26&until=2022-02-26&failure=false&probe_cc=AT&test_name=facebook_messenger)
  2. Open Console from there and you can see the console log
  
## Specifications

  - Version: Windows 11
  - Platform: Windows

majakomel:
Resolved in https://github.com/ooni/explorer/pull/710

PR #710, opened 2022-03-18
fix: remove console logs in the explorer

thara003:
## Checklist

- [x] reference issue for this pull request: [https://github.com/ooni/explorer/issues/680](https://github.com/ooni/explorer/issues/680)

## What kind of change does this PR introduce?

This PR ensures that there no more console logs in the production server


scripts/build-translations.js
```diff
@@ -26,4 +26,4 @@ const translationsMap = supportedLanguages
 
 const translationsContent = `window.OONITranslations = ${JSON.stringify(translationsMap)}`
 writeFileSync(`${LANG_DIR}/translations.js`, translationsContent)
-console.log(`> Wrote translations to: ${LANG_DIR}/translations.js`)
```
hellais, line 29: I think we should be keeping these `console.log` lines inside of the scripts, since it's expected for them to print some output to the user of them.

scripts/gen-flags.js
```diff
@@ -26,9 +26,7 @@ const main = () => {
         const jsContent = Mustache.render(tmpl, {alpha2, alpha2Upper})
         const dstPath = path.join(__dirname, '..', 'components', 'flags', `${alpha2}.js`)
         fs.writeFile(dstPath, jsContent, (err) => {
-          if (err) {
-            console.log('Failed to write ' + dstPath, err)
```
hellais, line 30: It's probably also a good idea to log this error in the script usage.

scripts/generate-csv.js
```diff
@@ -20,4 +20,4 @@ const defaultMessages = glob.sync(`${LANG_DIR}.messages/**/*.json`)
   }, [])
 
 writeFileSync('./static/lang/en.csv', JSON.stringify(defaultMessages, null, 2))
-console.log(`> Wrote default messages to: "${resolve('./static/lang/en.csv')}"`)
```
hellais, line 23: Same as for the above comments.

components/utils/profiler.js
```diff
@@ -2,7 +2,7 @@ import { Profiler as NativeProfiler } from 'react'
 
 export const Profiler = ({ id, children }) => {
   const _id = `${id || children.type.type.displayName}.${children.key}`
-  // console.log(children)
```
hellais, line 5: This and the above console.log can probably be removed as it looks like they are here just for debugging. But I will let @sarathms comment on if he considers them useful to keep.

scripts/build-translations.js
```diff
@@ -13,7 +13,7 @@ const supportedLanguages = glob.sync(`${LANG_DIR}/*.json`).map((f) => basename(f
 
 // Copy latest files from `translations`
 supportedLanguages.forEach((lang) => {
-  console.log('> Getting latest translations for langugae ✨', lang)
```
hellais, line 16: I think this should be kept, because it's part of the CLI tool output.

hellais (approved):
LGTM

PR #711, opened 2022-03-20
Replace react-datetime with react-day-picker

majakomel:
Closes https://github.com/ooni/explorer/issues/689

I only replaced the date picker on`/search` page and would be happy to hear feedback before applying changes to the other forms.

I decided to use [react-day-picker](https://react-day-picker.js.org/) because it's very customizable and it doesn't depend on any date library.

After the changes are applied to other forms as well, moment.js can be safely removed from the project.


sarathms (commented):
@majakomel Thanks for rewriting the date picker. Important change.

I haven't finished reviewing everything. However, I thought I should share some early feedback for us to discuss.

1. Looks like a new major version ([v8](https://github.com/gpbl/react-day-picker/releases/tag/v8.0.0)) of `react-day-picker` was released recently. And the documentation for v7 says that it [won't be maintained anymore](https://react-day-picker-v7.netlify.app/). I wonder if it is better we start out with v8 which seems to have some useful improvements over v7 like [reduced size](https://bundlephobia.com/package/react-day-picker@8.0.0), [tree-shaking](https://bundlephobia.com/package/react-day-picker@7.4.10), [hooks API](https://react-day-picker.js.org/api/functions/useDayPicker)
2. It does depend on the date library `date-fns` because it is a peerDependency. This makes the use of `dayjs` as replacement (#699) for `moment` a bit redundant. If we go ahead with `react-day-picker`, we might as well just use `date-fns` as the replacement for momentjs. It seems to be a well written library, except that it is not a drop-in replacement for `moment`. @hellais what do you say?
3. There seem to be merge conflicts after #699 was merged because this PR also adds a different version of `dayjs`. This might need a fresh rebase for this branch.


Looking at the implementation, I feel it might benefit from the hooks-based API of `react-day-picker@8.0.0`
* We can possibly avoid using the `FormProvider` and `useFormContext` route of implementing the search form
* Based on above point, the `DateRangePicker` component can be implemented independent of the `react-hook-form` library which allows us to use it flexibly in other places if needed.

majakomel:
@sarathms Thanks for checking the PR and for the tips for improvements!

I was so consumed with trying to find a nice date picker alternative, that I completely missed the newer version. 
I agree that going with v8 makes much more sense, just a bit of a bummer this one is dependent on `date-fns`. Otherwise, I really like the documentation and flexibility it offers.

I'll first upgrade the `react-day-picker`, after that I'll try to find an elegant solution to deal with UTC dates (day.js/moment.js offer a nicer way to do it than `date-fns`), in order not to carry two different date libraries.

majakomel:
@hellais I have added the styling change.
The only thing here in order to get completely rid of `react-datetime` component is to figure out what to do with /experimental/website, which is already a bit broken. Is this something we want to maintain?

hellais (approved):
LGTM!

We can drop /experimental/websites either as part of this PR or future work.

PR #712, opened 2022-03-21
Replace normal props with Transient props

girishsontakke:
This is a continuation of #560 which deals with the #487 issue.


sarathms (commented):
@girishsontakke Thanks for working on this.

I believe most of these props do not cause the warnings mentioned in #487. Can you please double check if there were warnings in the console (browser and node) that these changes address? 

Here's an example of what those warnings look like:
image

PR #720, opened 2022-03-30
Polish MAT form

sarathms:
- Use readable labels in MAT form
- Show some filters conditionally
- Limit date selection to 1 year duration
- Add subtitle to the page
- Update translations
- Shows a confirmation dialog when selected date range is more than 12 months (not field validation)
- Move x-axis and y-axis to first row


hellais:
Ah OK I see now that some of these issues are addressed in #724. Can you please merge all the branches together so I can review them all at once? Because some things are working in one branch, but not in the other.

sarathms:
All these changes were included in #724. Closing.

PR #721, opened 2022-03-31
MAT URL changes

sarathms:
- Move MAT to /chart/mat
- Add basic E2E test to check if MAT loads

### DIFF
--- a/cypress/integration/mat.e2e.js
+++ b/cypress/integration/mat.e2e.js
@@ -1,0 +1,19 @@
+describe('MAT Tests', () => {
+
+  describe('MAT redirections', () => {
+    it('Old /experimental paths redirect to /chart/mat', () => {
+      cy.visit('/experimental/mat?probe_cc=IR&test_name=web_connectivity&category_code=NEWS&since=2022-03-01&until=2022-04-01&axis_x=measurement_start_day&axis_y=probe_asn')
+      cy.url().should('include', '/chart/mat')
+    })
+  })
+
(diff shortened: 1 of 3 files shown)


next.config.js
```diff
@@ -13,6 +13,16 @@ const SentryWebpackPluginOptions = {
 }
 
 module.exports = withSentryConfig({
+  async redirects() {
+    return [
+      {
+        source: '/experimental/mat',
+        destination: '/chart/mat',
+        permanent: true,
```
hellais, line 21: It's worth noting that this uses the 308 status code instead of 301 to implement the permanent redirect.

From what I can tell this is supported in most modern browsers, so it should be fine to use. Search engines also seem to prefer it over 301.

hellais (approved):
LGTM

cypress/integration/mat.e2e.js
```diff
@@ -0,0 +1,19 @@
+describe('MAT Tests', () => {
+
+  describe('MAT redirections', () => {
+    it('Old /experimental paths redirect to /chart/mat', () => {
+      cy.visit('/experimental/mat?probe_cc=IR&test_name=web_connectivity&category_code=NEWS&since=2022-03-01&until=2022-04-01&axis_x=measurement_start_day&axis_y=probe_asn')
+      cy.url().should('include', '/chart/mat')
```
hellais, line 6: Maybe you can also add checks in here to see if the URL path also includes one of the query parameters?

PR #723, opened 2022-04-01
Chart fixes

sarathms:
- Use readable lable in MAT form
- Show some filters conditionally
- Limit date selection to 1 year duration
- Add subtitle to the page
- Update translations
- Use smaller heading for subtitle
- Show a confirmation screen before submission
- Show confirmation when the duration is longer than a defined duration
- Move x-axis and y-axis form elements to first row
- Remove undefined parts in tooltip titles
- Move OONI Logo to standard position on charts
- Remove legend from 1d chart


sarathms:
Replaced by #724

PR #724, opened 2022-04-01
MAT: Improve table-chart interaction

sarathms:
- Remove undefined parts in tooltip title
- Move OONI Logo to standard position on charts
- Remove legend from 1d chart
- Refactor GridChart to simplify data structures used for charts and table


hellais:
In here many of the issue reported in #720 are fixed.

Other issues spotted, that I don't know if they would be fixed if these branches are merged together, include:
- When I have applied a filter that I derived by searching, when I click the reset button the table view is reset, yet the ticks are not reset


https://user-images.githubusercontent.com/424620/161601748-ac979729-912a-480d-b500-ae8fc64bf275.mov

hellais:
I can confirm that the above reported issues have been resolved.

Following more testing this is what I have found:
- The tooltip is not visible in the 2d charts when a very low row count is chosen (see screenshot):
Screenshot 2022-04-05 at 14 42 38

-  When you resize the chart view, the visualization does not grow:

Screenshot 2022-04-05 at 14 44 05

- When you choose as an x-axis probe_asn the page crashes, see: https://explorer-git-mat-chart-fixes-ooni1.vercel.app/chart/mat?probe_cc=DZ&test_name=web_connectivity&since=2022-03-06&until=2022-04-06&axis_x=measurement_start_day&axis_y=probe_asn

- When no search filter is applied to the table and you select some items, after you click on reset, the chart updates, but the old ticks remain selected:

https://user-images.githubusercontent.com/424620/161756718-a300b712-f547-4abb-9c67-1d4112a7d040.mov

- The link to the measurements for the 1d chart should open in a new window, like the in the 2d charts (ideally it would be using the same tooltip with a link item, but we can defer this to post-launch)

PR #734, opened 2022-04-10
Add amount of measurements collected per country per year

### LINKED ISSUE #266: Display amount of measurements collected per country per year
Currently on OONI Explorer, you can only see how many measurements have been collected per country over the years. 

**Community members have expressed the need to be able to view how many measurements have been collected from each country per year.** 

It may be worth considering adding support for a feature which allows users to filter the measurements per country based on varying time-frames (year, month, week, etc.).

minhducsun2002:
Closes #266.

### DIFF
--- a/pages/countries.js
+++ b/pages/countries.js
@@ -9,7 +9,8 @@ import {
   Flex, Box,
   Container,
   Link, Input,
-  Heading, Text
+  Heading, Text,
+  Button
 } from 'ooni-components'
 import { StickyContainer, Sticky } from 'react-sticky'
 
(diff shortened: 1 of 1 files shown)


hellais:
Thanks for putting this PR together!

There seems to be an issue with the deployment and the tests not passing.

I'm also not 100% convinced of the general approach taken in here.

Since you are putting the fetching of the stats inside of `getInitialProps`, the fetching of these resources is going to happen as part of the server-side rendering, which is going to have an impact on the first page load time.

I think we ought to be doing this fetch client-side only, perhaps with a collapsable/expandable button on each of the country stat cards.

PR #737, opened 2022-04-11
Show country-wise data for twitter.com when MAT opens from NavBar

sarathms:
This sets the default chart shown on `/chart/mat` to "country-wise daily-counts for `twitter.com` in last 30 days"


hellais (approved):
LGTM

sarathms:
Limiting the chart configuration to 'All Web Connectivity data in the last 30 days' as discussed with @agrabeli.

PR #740, opened 2022-04-12
Handle when invalid axis options cause failures

sarathms:
- Do not allow x-axis to be empty
- Show friendly message when API request or data parsing fails

image


hellais (approved):
LGTM

Issue #745, opened 2022-04-15
Improve developer doc 


thara003:
# Improve documentation

## Link

[<https://github.com/ooni/explorer#readme>](<https://github.com/ooni/explorer#readme>)

## Describe the problem

In the perspective of a new contributor, the `README.md` provided seems to be very basic for setting up the project. So it would be good, if we have a `CONTRIBUTING.md` that covers all aspects of setting up the project locally and contributing.

## Describe the improvement

I think we should have `CONTRIBUTING.md` doc for the project and few things that we can include are:

- pointer to OONI Software Development Guidelines
- any helpful resources for getting started with the project
- specify how to open an Issue or Pull Request using one of our templates
- mention how to start contributing
- direction to reach us, if they have trouble in the setting up of the project

these are things that i have currently in my mind and it can extend to other project oriented things as well.

Issue #739, opened 2022-04-12
yarn run dev doesn't work on windows


hellais:
Running yarn run dev on windows leads to problems because the windows shell does not understand `.`.

Specifically the error message is as follows:
```
$ . ./scripts/populate-git-env.sh && next dev -p 3100 '.' is not recognized as an internal or external command, operable program or batch file. error Command failed with exit code 1. info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I guess a possible workaround for windows developers, would be to use [WSL](https://docs.microsoft.com/en-us/windows/wsl/install), but we should also consider making it work out of the box without needing these customisations.

Effectively we are using this bash script only for injecting the values of `GIT_COMMIT_*`. What we do is that we run `git` commands and set the output to environment variables and then inject these values into the app through the `webpack.DefinePlugin`.

I wonder if we can bypass the enviroment variable setting altogether and just call git directly from inside the `next.config.js` file as part of `webpack.DefinePlugin`: https://github.com/ooni/explorer/blob/master/next.config.js#L29

harisankar01:
Changing the path in script can help. Can add this to [Readme.md](https://github.com/ooni/explorer#readme) as a note for programmers using  windows as there is already an issue regarding this which was closed after giving the instructions about changing the path of scripts like issue https://github.com/ooni/explorer/issues/725. I will make a pull request regarding this. Thank You. The correct packages.json file for windows:
```
{
  "name": "ooni-explorer",
  "version": "2.9.0",
  "author": "Open Observatory of Network Interference (OONI) <contact@openobservatory.org>",
  "license": "BSD-3-Clause",
  "main": "index.js",
  "repository": "https://github.com/ooni/explorer",
  "dependencies": {
    "@babel/core": "7.4.4",
    "@datapunt/matomo-tracker-react": "^0.3.1",
    "@nivo/bar": "^0.79.1",
    "@nivo/calendar": "^0.79.1",
    "@nivo/core": "^0.79.0",
    "@nivo/funnel": "^0.79.1",
    "@nivo/heatmap": "^0.79.1",
    "@rebass/forms": "^4.0.6",
    "@sentry/nextjs": "^6.13.3",
    "axios": "^0.21.1",
    "babel-plugin-inline-react-svg": "1.1.0",
    "babel-plugin-styled-components": "^1.10.7",
    "buffer-from": "^1.1.1",
    "country-util": "^0.2.0",
    "dayjs": "^1.10.8",
[...]

PR #747, opened 2022-04-17
Updating ReadMe to support Windows user

harisankar01:
As per https://github.com/ooni/explorer/issues/739, I have added details required for windows developers to run the yarn dev on local environment without errors by changing scripts path in `package.json` and also added extra documentation to support new users (#745).


sarathms:
Hi @harisankar01. Thanks for thinking about how to support development on Windows. However, the change you proposed here needs to be discussed further. There might be more scenarios, but consider the situation when we make changes the `scripts` section in `package.json`. Now we have to remember to make parallel changes in this section of `Readme.md` which is very likely to be missed and then they are out of sync.

It might create confusion when these changes are accidentally committed by someone as part of other code changes.

As an alternative, can you use separate entries in `scripts` to be used in windows? e.g
````
   "build:win": "a different command for windows",
   "test:win": "...",
````

**Comment by harisankar01** [comment]
Okay. Let me work on it.

**Comment by harisankar01** [comment]
Made the changes as per the instructions. Also has issue #746 . Thank you

File: package.json [hunk 1]
```diff
@@ -78,10 +78,14 @@
...
   "scripts": {
     "dev": ". ./scripts/populate-git-env.sh && next dev -p 3100",
+    "dev:win": "./scripts/populate-git-env.sh && next dev -p 3100",
     "debug": ". ./scripts/populate-git-env.sh && WDYR=1 NODE_OPTIONS='--inspect' next dev -p 3100",
+    "debug:win": "./scripts/populate-git-env.sh && WDYR=1 NODE_OPTIONS='--inspect' next dev -p 3100",
     "start": "next start -p 3100",
     "build": ". ./scripts/populate-git-env.sh && next build",
+    "build:win": "./scripts/populate-git-env.sh && next build",
     "export": ". ./scripts/populate-git-env.sh && next export",
+    "export:win": "./scripts/populate-git-env.sh && next export",
```
hellais, line 88: These two commands are not really equivalent to each other, since you aren't running the shell script with with `.` (or `source`), the exported environment variables aren't actually being included in the context of `next *`.

I think we either find a way to make the definition of these env vars work on windows, or we just drop the running of those scripts on windows so that we don't mislead a developer into believing they are actually doing something.

pages/search.js
```diff
@@ -338,7 +340,7 @@ class Search extends React.Component {
           // When `hideFailure` is true, add `failure=false` in the query
           query[queryParam] = false
         } else {
-          delete query[queryParam]
+          query[queryParam] = true
```
hellais, line 343: I think this change came from a different branch and should be included a separate PR.

harisankar01: Okay let me change it.

Issue #744, opened 2022-04-14
Create network centric views for Explorer


hellais:
This is about having pages, that live under `/network/{asn}` which provide information about blocking from the perspective of a particular network.

I think we should have these pages be broken down in sections and for each section we can create the relevant MAT chart.

Here is a quick mockup of how this could look like:
![Desktop HD.png](https://images.zenhubusercontent.com/58442543b9dc3c622022e97d/25ea6f32-f7ff-4a9f-b0c7-19bf3a4f2988)

hellais:
For the high level stats areas, I think the bits of information we should include are:
- Total number of measurements collected from that network since the beginning of time
- Date of the first measurement

In addition to this I would include a coverage chart (in calendar form), similar to what we plan to have added to the country pages. See this comment for details: https://github.com/ooni/explorer/issues/687#issuecomment-1069053503

PR #750, opened 2022-04-25
New network centric view

majakomel:
Closes https://github.com/ooni/explorer/issues/744


majakomel:
@hellais This is a rough implementation of the network view - it needs some more refactoring/cleanup and changes to the date selector when the other ticket is ready to be merged.

Is there any clear idea yet what to put on the top of the page in the "high-level stats" area? I can start working on that part too.

hellais:
Thanks for putting this together. I left in the original issue some ideas for what we can do with the high level stats: https://github.com/ooni/explorer/issues/744#issuecomment-1108467317.

Re: this branch, here is some feedback:

1. Because of how the API works, asking for `AS1234` is the same as asking for `1234`. I think on these pages, though, we should enforce stricter checks on what is an allowed ASN (we should ensure it starts with `AS`). Not doing so will lead to two pages having the same content which, amongst other things, is bad for SEO.
2. We should have a better page for when a network does not exist. I would say we can detect when we have no data for a particular network and display something like "This network either does not exist or we have no data for it"
3. The AS number should be displayed prominently at the top of the page

components/network/MessagingAppsChart.js
```diff
@@ -0,0 +1,133 @@
...
+      console.log(e)
+      e.message = e?.request?.response ?? e.message
+      throw e
+    })
+  }
+
+  const signal = request(`${reqUrl}&test_name=signal`)
+  const telegram = request(`${reqUrl}&test_name=telegram`)
+  const whatsapp = request(`${reqUrl}&test_name=whatsapp`)
+  const fbMessenger = request(`${reqUrl}&test_name=facebook_messenger`)
```
hellais, line 46: This is a good interim solution for this, though we ought to really be able to do this with a single request to the backend, see: https://github.com/ooni/backend/issues/584

components/network/CircumventionToolsChart.js
```diff
@@ -0,0 +1,125 @@
...
+      return {
+        data: r.data.result,
+        loadTime: r.loadTime,
+        url: r.config.url
+      }
+    }).catch(e => {
+      e.message = e?.request?.response ?? e.message
+      throw e
+    })
+  }
```
hellais, line 38: I see this is used in both the IM app page and this one, perhaps this can be refactored into some shared function?

components/network/CircumventionToolsChart.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const circumventionTools = ['psiphon', 'tor', 'torsf']
+  const requests = circumventionTools.map((tool) => request(`${reqUrl}&test_name=${tool}`))
+
+  return Promise.allSettled(requests).then((responses) => {
+    return responses.reduce((prev, current, index) => [...prev, ...current.value.data.map((obj) => ({...obj, circumvention_tool: circumventionTools[index]}))], [])
+  })
+}
```
hellais, line 46: In fact, the whole fetcher function could probably be refactored to work for both the IM apps and the circumvention apps.

pages/network/[asn].js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const fixedQuery = {
+  axis_x: 'measurement_start_day',
+}
+
+const testsOnPage = ['web_connectivity']
+const ChartsContainer = () => {
+  return (
+    testsOnPage.map(testName => (
```
Comment (hellais) [inline pages/network/[asn].js:30]: Why are is map being called on a list that has a single element? Isn't this equivalent to not having it being called at all?

pages/network/[asn].js
```diff
@@ -0,0 +1,86 @@
...
+      <MATContextProvider key={testName} test_name={testName} {...fixedQuery}>
+        <WebConnectivityChart testName={testName} />
+        <MessagingAppsChart testName='messaging_app' />
+        <CircumventionToolChart testName='circumvention_tool' />
```
Comment (hellais) [inline pages/network/[asn].js:34]: I suspect these three different chart types can be refactored so that there is significantly less duplication between them.

majakomel:
@hellais I refactored all charts into one component and extracted two fetchers. Should I add a TODO comment or is there a way to add a follow-up ticket when backend adds support for multiple test_names?

Few notes: 
- I downgraded `@nivo/core` until the v0.79.1 with regression fixes is published (https://github.com/plouc/nivo/issues/1928)
- For now I didn't add custom ranges - the colors are distributed based on how the library works by default (calculated based on min/max of the whole data range). There seems to be the functionality to define a custom range, but it's not documented so I have to yet figure it out.
- Can you give me some more background on the scrollbar issue on the web connectivity chart? Currently it looks the same as on MAT page, but maybe I'm missing something.

utils/index.js
```diff
@@ -45,3 +45,8 @@ export const truncateString = (s, maxStart, maxEnd) => {
   truncatedString += s.substr(s.length - maxEnd, s.length)
   return truncatedString
 }
+
+export const getRange = (start, end) => {
+  if(start === end) return [start]
+  return [start, ...getRange(start + 1, end)]
```
hellais, line 51: I think this function would be better rewritten as an iterative one. Different javascript engines have different limits on the maximum level of recursion.

For example with node v16.8 I already hit the limit when using a range of 9000.

An possible iterative implementation could look something like:
```
const getRangeLoop = (start, end) => {
  if(start === end) return [start]
  return [...Array(end - start + 1).keys()].map(idx => idx + start)
}
```

I would suggest adding also some guards to ensure that end is > start or return just start in that case.

pages/network/[asn].js
```diff
@@ -0,0 +1,131 @@
...
+    <Layout>
+      <MetaTags />
+      <NavBar />
+      <Container>
+        <Heading h={1} fontWeight='heading' my={20}>AS{asn}</Heading>
+        {router.isReady && 
+          <React.Fragment>
+            <Text fontSize={20} my={10}>Total number of measurements: <Bold>{measurementsTotal}</Bold></Text>
+            <Text fontSize={20} my={10}>Date of the first measurement: <Bold>{calendarData[0].day}</Bold></Text>
+            <Text fontSize={20} my={10}>Network observed in countries: <Bold>{countries.map(c => countryUtil.territoryNames[c]).join(', ')}</Bold></Text>
```
Comment (hellais) [inline pages/network/[asn].js:76]: These strings should use `FormattedMessage` so that they can be translated

hellais:
This is looking good.

Some comments based on an inspection of the preview:

1.

Screenshot 2022-05-16 at 16 26 39

In the measurement overview screen I think we should color code the dates in the future differently, since it's expected for there to not be any data there. I would suggest using an even lighter grey for those dates.

2.

For the dates which have zero data, it would be good if we had in those cases also a tooltip that shows 0 in it, so it's clear that is what is going on here.

3.

For the "Network observed in countries: XXX", I think we should make the XXX country name a link to the relevant country page so that we connect these two pages together.

4.
When I access a network for which we don't have any data, I think we should have a specific 404 page or have a zero state for that network page. We can display something similar to when we have no OONI data for a certain country encouraging people to run measurements: https://explorer.ooni.org/country/AQ.

5.

I think the URLs for the networks should start with AS{NUMBER}, this gives us the flexibility to re-route endpoints based on the prefix if we ever decide to change the format (maybe we want to put the network name in the address bar instead of the ASN).

Sorry if my initial comment on this point wasn't clear.

Answering your questions in the above comment:
> Should I add a TODO comment or is there a way to add a follow-up ticket when backend adds support for multiple test_names?

I would say we should document it in an ooni/explorer issue and link it to the backend one so we remember to do it once there is backend support.

> For now I didn't add custom ranges - the colors are distributed based on how the library works by default (calculated based on min/max of the whole data range). There seems to be the functionality to define a custom range, but it's not documented so I have to yet figure it out.

I think it would be good if we could figure out how to get the ranges to work properly. The issue with using ranges that are relative to a given network is that it will make it hard to compare charts of different networks (for example what is colored green in one network, might be colored red in another, because they have big spikes of many more measurements).

> Can you give me some more background on the scrollbar issue on the web connectivity chart? Currently it looks the same as on MAT page, but maybe I'm missing something.

I tested this again with the latest deployed version and I don't seem to be able to reproduce this anymore.

hellais:
Another thing that would be good to have, is if we could sort the list of "Network observed in countries" by the number of measurements and perhaps also display the total number of measurements for each country.

This is useful because it would allow us to spot cases in which the network is wrongly geolocated to a specific country.

utils/index.js
majakomel: Thanks for this, wasn't aware of it.

majakomel:
I addressed all the points from the comments, color range is also set now. Empty network page only shows CTA - let me know if you want to add more content.

hellais:
I think this looks good and we should merge and deploy it as is.

We should document some items to be done for future work, which can be implemented iteratively as follow up PRs:
* Move the loading of the calendar data into the client side. Currently the first load for some very big networks will take even in the range of seconds. We should instead be doing this client-side and display a loader while that is being done.
* Fetch the calendar data lazily. There is no need to prefetch all the calendar data until the over has opened it up. The only metadata we need to populate that section is the data of the first measurement. We can do this as a dedicated API endpoint and while we are at it we could also have it support returning the total number of measurements, so we don't have to do 2 requests.
* Link the network pages from other pages. We should add links to the network page in every measurement page as well as the country pages. This way they will be discoverable and exposed to end users.
* Create an index/search of all networks. We should have some way to do discovery and search of all the measured countries. This should go hand in hand with rethinking of the country page listing.

Issue #660, opened 2022-02-07
measurement page: measurement that shows evidence of blocking is rendered as accessible


hellais:
The following measurement presents signs of blocking: https://explorer.ooni.org/measurement/20180729T190855Z_AS30985_P80GekW4UhMA9BkXQKZ4WSSlSdAxuYGHuC2Aljz5r4SUIXBnVs, but it's rendered in OONI Explorer as accessible.

It also looks like the color of the headers is inconsistent with the accessible status, so there is something odd happening.
![Screenshot 2022-02-07 at 17 53 06](https://user-images.githubusercontent.com/424620/152834188-b0e5133e-c2b2-4a41-a1d0-45f7d141b98c.png)

PR #753, opened 2022-05-03
Fix text on WhatsApp measurement page when it's blocked

majakomel:
I used a similar logic to how determining anomaly is done for Signal and some other messaging components, the rest seems to work correctly now. 

Closes https://github.com/ooni/explorer/issues/660

### DIFF
--- a/components/measurement/nettests/WhatsApp.js
+++ b/components/measurement/nettests/WhatsApp.js
@@ -21,8 +21,10 @@ const BugBox = styled(Box)`
   padding: 20px;
 `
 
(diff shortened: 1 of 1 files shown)


sarathms:
Thanks @majakomel for looking into this. In general, we have been trying to move away from the approach of Explorer trying to determine the OK/Anomaly status of a measurement by looking into the `test_keys` and instead let the backend determine this. It is true that we still do this in some cases, but the idea is to eventually not do those either.

For this particular case, issue there is a backend issue (ooni/backend#559) that when addressed should fix this issue without any changes to Explorer. It should have been documented in the issue this PR intends to address. Sorry about that.

Issue #701, opened 2022-03-14
measurement page crashes when accessing missing measurement from search interface


hellais:
When I attempt to access a missing measurement from the search listing, the page crashes with an error:
![Screenshot 2022-03-14 at 18 11 10](https://user-images.githubusercontent.com/424620/158224646-6e80b5dc-00a2-4dd8-96ff-662168c1b503.png)

When doing the same thing, but not by navigating to the view, I instead get a missing measurement page as expected:
![Screenshot 2022-03-14 at 18 11 16](https://user-images.githubusercontent.com/424620/158224715-17556be5-7d67-41a8-af11-6639f5677a92.png)

PR #769, opened 2022-07-12
Fix parsing error when there is no raw_measurement

majakomel:
Also closes https://github.com/ooni/explorer/issues/701


hellais (approved):
LGTM!

PR #780, opened 2022-08-08
chart header responsive legend

GermaVinsmoke:
Current version of legend looks like this on mobile which is not responsive at all. The legends are overlapping on each other.

image

Changed it to this version 

image


hellais (approved):
LGTM

GermaVinsmoke:
This was done only in one chart but then I saw that other charts are also having this problem, so I guess I'll do the changes in those also.

Issue #771, opened 2022-07-25
coverage chart: change color palette to the ooni blue scale


hellais:
Change the color of the coverage calendar chart to something else, we could do it ooni blue. Otherwise it's misleading that it looks like the blocked/anomaly indicators, but instead it's talking just about volume of coverage.

We should also add a color scale at the bottom of the chart.

Issue #424, opened 2020-03-16
Expose confirmed blocked category codes in the country page


hellais:
As a simple PoC we could have:
* Categories of websites confirmed blocked
* Bool flag to indicate if IM apps are blocked, if middleboxes were detected and if circumvention tools are restricted

This also requires some data analysis from @FedericoCeratto and a mockup from @hellais. 

This is also related to: https://github.com/ooni/explorer/issues/386

hellais:
I went ahead and made a mocked implementation for how this should work in the following branch: https://github.com/ooni/explorer/tree/confirmed-categories.

Here a screenshot of how this looks:
![Screenshot 2021-04-19 at 14 56 25](https://user-images.githubusercontent.com/424620/115239949-8efdfc00-a11f-11eb-948a-cbceefb7fc7b.png)

I am going to now unassign this issue from myself and leave it assigned to @sarathms for finalising the implementation.

hellais:
This should be possible by running the following query on the aggregation API: https://api.ooni.io/api/v1/aggregation?probe_cc=IR&since=2021-06-07&until=2021-07-08&test_name=web_connectivity&axis_x=category_code.

hellais:
We have this now

PR #785, opened 2022-08-11
replaced older IM and circumvention charts

GermaVinsmoke:


GermaVinsmoke:
Had to add a separate entry point `Chart.js` because the way query and query params are formed are different in the `Chart.js` of `network` page. I think if we can separate out the query generation logic and `useSwr` query hook into separate files then we can generalise the `Chart.js` file into a single `component`.

hellais:
This looks great, thanks for putting it together. Some things I think we should do before merging it are:

1. Remove the commented out parts of code and delete them
2. There are several components that are no longer needed and they should also be removed (e.g. `inCountry` context)
3. We need some sort of time interval picker for the charts on the country page (similar to what we have on the network page) so that the user is able to filter by a specified date range
4. Some bits of the page are not translatable (i.e. they need to be wrapped using the `FormattedMessage` component), I will add an inline comment for these
5. Since the country and network page charts are different, to avoid confusion I would give these two components a different name (e.g. `ChartCountry`, `ChartNetwork`) and eventually consider merging them as you said in your comment.

Next steps, which can be done as part of a separate PR:
1. Refactor the coverage chart to make use of the coverage chart we have in the network page (while you are at it I would also address: https://github.com/ooni/explorer/issues/771)
2. As you mentioned in your above comments, look into the possibility of sharing the Chart component between the country and network pages

Other nice to have improvements to the country page include:
* https://github.com/ooni/explorer/issues/424
* https://github.com/ooni/explorer/issues/266

GermaVinsmoke:
Hey @hellais , I did 1 and 2. Removed the unused components and files. I wanted to know what to do about some of the section headers, navigation and styling things.

1. The navigation present here, since we removed the Networks section at last, the scroll link doesn't work anymore.
image

2. The section headers, are those all good or can we improve that part? Because it still doesn't match - In the Websites section, we got 2 headers of Website.
Then the last 2 charts (Instant Messaging & Circumvention) are present inside Apps section.

hellais:
> The navigation present here, since we removed the Networks section at last, the scroll link doesn't work anymore.

Yeah I would say we just drop the networks tab until we re-add support for it.

> The section headers, are those all good or can we improve that part? Because it still doesn't match - In the Websites section, we got 2 headers of Website.

Good catch. I think we would ideally at the very least drop the Websites heading from the Websites section, so we aren't repeating it twice.

> Was thinking do we need to add the filter parameters in URL also?

Yeah I think that's a solid plan

components/aggregation/mat/timeScaleXAxis.js
```diff
@@ -5,10 +5,12 @@ import { getDatesBetween } from './computations'
 
 const defaultCount = 20
 
-export function getXAxisTicks (query, count = defaultCount) {
+export function getXAxisTicks (data, count = defaultCount) {
 
-  if (query.axis_x === 'measurement_start_day') {
```
majakomel, line 10: We still need this condition, otherwise the charts with `query.axis_x` other than `measurement_start_day` are broken. With moving query to params the changes here will also not be necessary anymore.

majakomel: I'm referring to your question and Arturo's answer about having parameters in the URL. In that case, this change won't be necessary anymore.

> > Was thinking do we need to add the filter parameters in URL also?
> 
> Yeah I think that's a solid plan

GermaVinsmoke: Hey @majakomel @hellais
Added date and form in the URL of the country page.

GermaVinsmoke: So, now the `ChartCountry` and `ChartNetwork` components are the same, if everything looks good then I can convert those two back to one file.

components/country/ChartCountry.js
```diff
@@ -0,0 +1,96 @@
...
+
+  const headerOptions = { probe_cc: false, subtitle: false }
+
+  return (
+    <MATContextProvider key={name} test_name={name} {...params}>
+      <Flex flexDirection='column' mb={60}>
+        <Box><Heading h={3} mt={40} mb={20}>{title}</Heading></Box>
+        <Box>
+          {(!chartData && !error) ? (
+            <div> Loading ...</div>
```
hellais, line 65: This needs to be wrapped with `FormattedMessage`

components/country/ChartCountry.js
```diff
@@ -0,0 +1,96 @@
...
+          ) : (
+            chartData === null || chartData.length === 0 ? (
+              <Heading h={5}>No Data</Heading>
```
hellais, line 68: This needs to be wrapped with `FormattedMessage`

components/country/ChartCountry.js
```diff
@@ -0,0 +1,96 @@
...
+                isGrouped={false}
+                header={headerOptions}
+              />
+            )
+          )}
+        </Box>
+        {error &&
+          <DetailsBox collapsed={false} content={<>
+            <details>
+              <summary><span>Error: {error.message}</span></summary>
```
hellais, line 83: This needs to be wrapped with `FormattedMessage`

PR #788, opened 2022-08-17
Stricter report_id validation

majakomel:
Adds stricter report_id regex to avoid making unnecessary api requests.
Regex matches the one on the backend: https://github.com/ooni/api/pull/301

### DIFF
--- a/pages/measurement/[[...report_id]].js
+++ b/pages/measurement/[[...report_id]].js
@@ -40,7 +40,7 @@ export async function getServerSideProps({ query }) {
   // in which case, the extra segments are available inside query.report_id[1+]
   const report_id = query?.report_id?.[0]
   // If there is no report_id to use, fail early with MeasurementNotFound
-  if (typeof report_id !== 'string' || report_id.length < 1) {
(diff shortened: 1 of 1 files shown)


pages/measurement/[[...report_id]].js
```diff
@@ -40,7 +40,7 @@ export async function getServerSideProps({ query }) {
   // in which case, the extra segments are available inside query.report_id[1+]
   const report_id = query?.report_id?.[0]
   // If there is no report_id to use, fail early with MeasurementNotFound
-  if (typeof report_id !== 'string' || report_id.length < 1) {
+  if (typeof report_id !== 'string' || !report_id.match(/[a-zA-Z0-9_-]{5,100}/)) {
```
Comment (hellais) [inline pages/measurement/[[...report_id]].js:43]: I think we can be even a bit more aggressive with the filter here by setting the minimum length to 40 (according to the fastpath table the shortest non-null report_id is 45 characters, and it's of the new format `20201028T073929Z_ndt_ZZ_0_n1_Ed95brGpL3sZ4dUN`, so even in the case of a single character test name we would be at 43 chars).

hellais (approved):
Suggested a small improvement to the report_id validation filter, but otherwise LGTM

Issue #779, opened 2022-08-08
view measurements link in MAT should include failed measurements if failure_count is not 0


anadahz:
## Expected Behavior

The `view measurements link` in MAT should list failed measurements if `failure_count` is not 0. 

## Actual Behavior

The `view measurements link` in MAT is not listing failed measurements if when `failure_count` is greater than 0. By default the `Hide failed measurements` option is checked on. This is problematic for days/countries that present only failed measurements, in this cases the `view measurements link` will reveal no measurements.  

## Steps to Reproduce the Problem

  1. Use MAT (https://explorer.ooni.org/chart/mat)
  2. Click on `view measurements` on a day that includes failed measurements

PR #789, opened 2022-08-18
Include failed measurements when linking from MAT

majakomel:
Closes https://github.com/ooni/explorer/issues/779


hellais (approved):
LGTM

Issue #774, opened 2022-08-01
Implement login functionality as the ground work for user feedback reporting system


majakomel:
It should use the same endpoint and behaviour as login action in Test List Editor.

Related ticket: https://github.com/ooni/ooni.org/issues/1200

PR #790, opened 2022-08-22
User feedback mechanism with login

majakomel:
Closes https://github.com/ooni/explorer/issues/774
Closes https://github.com/ooni/explorer/issues/811


components/measurement/FeedbackBox.js
```diff
@@ -0,0 +1,254 @@
...
+        }
+      },
+      success: {
+        on: {
+          CLOSE: 'closed'
+        }
+      },
+      failure: {
+        on: {
+          CLOSE: 'closed',
```
hellais, line 91: If I understand how this state machine works accurately, these on event handlers are those responsible for defining what the next state is in response to calling `send` when inside that particular state.

Specifically here in the `failure` state I don't see any call to the `send("CLOSE")`, so it's unclear to me why this is being handled.

components/measurement/FeedbackBox.js
```diff
@@ -0,0 +1,254 @@
...
+          const { nested, feedback } = getValues()
+          const feedbackParams = {
+            measurement_uid: report_id,
+            status: nested || feedback
+          }
+          submitFeedback(feedbackParams)
+            .then(() => send('SUCCESS'))
+            .catch(({response}) => {
+              setError(response?.data?.error || 'unknown')
+              send('FAILURE')
```
hellais, line 76: There is something odd with how the failures are being handled. I tried from the test instance to submit feedback, but the UI showed a successful submission, while according to the network inspector there was actually a CORS error.

components/measurement/FeedbackBox.js
```diff
@@ -0,0 +1,254 @@
...
+                        <QuestionText i18nKey='Measurement.Feedback.Q1' />
+                        <Button mr={2} onClick={() => send('IS_BLOCKING')}><FormattedMessage id='General.Yes' /></Button>
+                        <Button onClick={() => send('CLOSE')}><FormattedMessage id='General.No' /></Button>
+                      </>
+                    }
+                  </>
+                  : 
+                  <Text fontSize={16}>
+                    <FormattedMessage
+                      id='Measurement.Feedback.Login'
```
hellais, line 147: I think that we should only prompt the user to login once they have expressed interest in submitting feedback.

That is to say we should present the same initial state irrespective of the logged in state and only on the second step, only if they are not logged in, ask them to login/register to provide feedback.

pages/login.js
```diff
@@ -0,0 +1,90 @@
...
+  // useEffect(() => {
+  //   // if (!loading && user.loggedIn) {
+  //   //   router.replace('/')
+  //   // }
+  // }, [user, loading, router])
+
+  return (
+    <Layout title='Login'>
+      <Head>
+        <title>Login</title>
```
hellais, line 34: Maybe for the login page we should inherit the title tags coming from the rest of OONI Explorer and mention somewhere in here that it's a login relative to OONI Explorer.

hellais (changes requested):
It's looking good. I left some feedback for things to improve.

We also should also, before releasing it, make some improvements to the overall design of the page. At the very least use a color palette that is not greyscale.

components/measurement/FeedbackBox.js
```diff
@@ -22,29 +22,32 @@ const StyledError = styled.small`
...
 const blockedValues = [
   { top: 'ok' },
-  { top: 'ok.broken' },
-  { top: 'ok.parked' },
-  { top: 'ok.unreachable' },
-  { top: 'blocked',
-    sub: [
-  { 
-    top: 'blocked.blockpage',
+  { top: 'broken',
```
hellais, line 27: Following the discussion on slack, let's for the moment limit this to only 2 options:
* down.unreachable
* down.misconfigured

The goal is to capture under the `down` tree anything that has to do with issues with the service itself that are not attributable to censorship.

`unreachable` is meant to capture anything that renders the resource unreachable from anywhere in the world. It's probably useful to capture this separately, because in this instance, when the blocking is implemented in a way that is indistinguishable from the service being unreachable (ex. timeout), the outcome is ambiguous. There are still several cases in which we can say that blocking happening, even if the service is unreachable (ex. DNS blocks, HTTP proxies, RST injection, TLS MITMs).

`misconfigured` is any form of misconfiguration that renders the resource still available, yet it's not in working state. This can include instances where the domain is parked, it's serving a blank page or in any case not functioning as expected, while still being reachable.

hellais (approved):
LGTM! 🐳

PR #791, opened 2022-08-30
Migration/ooni components

GermaVinsmoke:


package.json
```diff
@@ -29,7 +28,7 @@
     "markdown-to-jsx": "^7.1.7",
     "next": "12.2.5",
     "nprogress": "^0.2.0",
-    "ooni-components": "^0.4.7",
+    "ooni-components": "^0.5.0-alpha.6",
```
hellais, line 31: I would say that before we land this, we go ahead with cutting a 0.5.0 release of ooni-components and bump the version tag in here to that.

GermaVinsmoke: Are we skipping the testing in other projects? 🤔
> we go ahead with cutting a 0.5.0 release

Are we creating a production release of ooni-components? Maybe we can also go with a beta tag release and add that in all the projects to see how it works with the build and deployed applications.

majakomel:
Closing as changes are no longer relevant.

PR #793, opened 2022-09-04
Fixed country search display message while no country was found

### LINKED ISSUE #792: Countries search - no result text
## Expected Behavior
It should show what we've typed in the search input field.

## Actual Behavior
image

Anish1008:
Fixes #792

- [x] Removed extra pair of curly braces around the "searchTerm" in countries page

### DIFF
--- a/pages/countries.js
+++ b/pages/countries.js
@@ -145,7 +145,7 @@ const NoCountriesFound = ({ searchTerm }) => (
         {/* TODO Add to copy */}
         <FormattedMessage
           id='Countries.Search.NoCountriesFound'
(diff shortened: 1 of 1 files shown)


majakomel (changes requested):
This is not the right approach, since we do need to pass an object to values prop.
It looks like a problem is rather with the translation string itself - we can either keep single quotes like this `"No countries found with ''{searchTerm}''"` or switch to double quotes `"No countries found with \"{searchTerm}\""`.

majakomel:
This has been meanwhile fixed in https://github.com/ooni/explorer/pull/796

Issue #413, opened 2020-02-26
Search queries should be done client side


hellais:
Currently when I access https://explorer.ooni.org/search, the queries are being performed server-side, instead these should be done client-side so that we don't block on rendering some content to the user.

This may also help solve the fact that while the search page is rendering it is not possible to toggle the search parameters.

PR #799, opened 2022-09-28
Search page refactoring

majakomel:
Closes https://github.com/ooni/explorer/issues/413

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -1,7 +1,7 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import Head from 'next/head'
-import { withRouter } from 'next/router'
+import { useRouter } from 'next/router'
 import axios from 'axios'
 import styled from 'styled-components'
 import {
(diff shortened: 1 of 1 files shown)


pages/search.js
```diff
@@ -143,292 +185,165 @@ const NoResults = () => (
...
-    }
-    if (this.state.loading != nextState.loading) {
-      return true
-    }
-    return false
-  }
+    setLoading(true)
+    getMeasurements(query)
+      .then((res) => {
+        const results = res.data.results.filter(item => item.probe_asn !== 'AS0')
```
hellais, line 209: I don't think we should be doing this sort of filtering inside of explorer. We ought to rather just display whatever the API is giving us.

pages/search.js
```diff
@@ -143,292 +185,165 @@ const NoResults = () => (
...
+      })
+  }, [])
 
-  loadMore() {
-    axios.get(this.state.nextURL)
+  const loadMore = () => {
+    axios.get(nextURL)
       .then((res) => {
-        // XXX update the query
         const nextPageResults = res.data.results.filter(item => item.probe_asn !== 'AS0')
```
hellais, line 225: Same comment as above and I now realise this was already here, but maybe it's an opportunity to do some cleanup of this too.

hellais (commented):
Overall this looks very good. I left a few small comments that can either be addressed now or in a following PR.

LGTM!

PR #801, opened 2022-09-30
open git version github link

GermaVinsmoke:


GermaVinsmoke:
@majakomel We discussed that since we're getting the commit hash value in the footer now, it'll be a good idea to open that link for anyone to check that version's code.

PR #806, opened 2022-10-10
adding git inside the container

GermaVinsmoke:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN yarn install --frozen-lockfile
 
 # Rebuild the source code only when needed
 FROM node:16.3-alpine3.12 AS builder
+RUN apk add --no-cache git
 WORKDIR /app
 COPY . .
 COPY --from=deps /app/node_modules ./node_modules


majakomel:
Just adding git is unfortunately not going to be enough with our current build process - there are other changes needed in the build script we use. 
I'm working on other build improvements here (still WIP) https://github.com/ooni/explorer/pull/808

GermaVinsmoke:
But when I tested with the current `DockerFile`, it was giving `git not found`.
On installing git inside the builder container, it was able to run those `yarn --silent` commands successfully 🤔

majakomel:
Yes, we need to add git, but it is not the only thing required for it to work, since we [currently run it against git archive](https://github.com/ooni/sysadmin/blob/master/scripts/docker-build), which is "not a git repository" (doesn't contain .git folder). I also found out there are some other optimizations we can do to drastically decrease docker image size.

GermaVinsmoke:
That means, even the previous `./git-populate-env.sh` script was wrong for populating the git env variables 🤔

majakomel:
Closing this since it's included in https://github.com/ooni/explorer/pull/808

PR #808, opened 2022-10-13
Build improvements

majakomel:
Followed the updated official example for latest next.js https://github.com/vercel/next.js/blob/canary/examples/with-docker

Decreases docker image size from ~1gb to less than 200mb

Previously:
Screenshot 2022-10-18 at 10 01 59

With changes:
Screenshot 2022-10-18 at 10 02 09


pages/about.js
```diff
@@ -10,23 +10,20 @@ import {
...
+      <NavBar />
 
+      <Container>
+        <Heading h={2}>XXX Implement Me</Heading>
+        <Text>Do we even need an about page?</Text>
+      </Container>
+    </Layout>
+  )
 }
+export default About
```
hellais, line 29: Should we maybe at this point just drop this page or setup a redirect to ooni.org/about?

Dockerfile
```diff
@@ -6,43 +6,41 @@
 # Based on this issue: https://github.com/docker/for-mac/issues/5831
 
 # Install dependencies only when needed
-FROM node:16.3-alpine3.12 AS deps
+FROM node:16-alpine AS deps
```
hellais, line 9: I'm not sure we want to be using the generic image name, it makes the build process more prone to potential issues and less stable.

If a new release of alpine breaks something, we can't easily tell which was the last known working version.

pages/about.js
majakomel: I have removed the about page, we didn't link to it anywhere and there is already a link to ooni.org/about in the footer.

Dockerfile
majakomel: Done

PR #809, opened 2022-10-17
Refactor MAT filters table to virtualize rows

majakomel:
Component with filters was split from the TableView components, row virtualization was added for filtering table.


hellais:
I did a very rough comparison of the performance profiles for the before and after:
Screenshot 2022-10-19 at 17 00 37

This was the URL I used to test: https://explorer.test.ooni.org/chart/mat?test_name=web_connectivity&since=2021-10-19&until=2022-10-19&axis_x=measurement_start_day&axis_y=domain.

It does seem like the memory usage goes down by a fair amount and the total number of nodes is down by almost an order of magnitude in my test-case.

Overall this seems like a very beneficial improvement.

One thing that we might want to look into though is the fact that on first load the second page of the table, when you scroll with the mouse, takes a fair amount of time to load.

I seem to recall there was the option to tell the virtualized row library to pre-load the next few pages worth of data so that you don't have such a noticeable lag on scrolling. It might be worth looking at what we do in other charts.

majakomel:
@hellais The 'overscan' setting is already set and it works fine for charts with shorter timespan - issue seems to be related to resource consuming rendering of charts, I'll try to find improvements for that in next iteration

PR #826, opened 2022-12-28
Add prettier

majakomel:


GermaVinsmoke:
How about setting some kind of styleguide also?

.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  semi: true,
```
GermaVinsmoke, line 2: Isn't the whole project discourages the use of **semi colon**?

chidexebere:
@majakomel  Please can this be revisited.

I have prettier installed on my code editor (I use VS code).
I had to disable prettier on my ooni explorer workspace because my default prettier setting was formatting the code.
There is a need for a prettier setting (`.prettierrc` file) to format the code as desired.

PR #855, opened 2023-05-19
Setup Prettier

chidexebere:
## Fixes  #826 

This:
- Installs `prettier` and `eslint-config-prettier` dependency
- Move the formatting rules to a newly created `.prettierrc` file
- Add files and folders prettier should ignore using a created `.prettierignore` file


majakomel:
Closing this since we're going with biome

PR #868, opened 2023-08-22
Use only backend provided values for vanilla_tor

hellais:
Fix checks for vanilla_tor to remove parsing of test_keys

### DIFF
--- a/components/measurement/nettests/VanillaTor.js
+++ b/components/measurement/nettests/VanillaTor.js
@@ -10,16 +10,7 @@ import { MdTimelapse } from 'react-icons/md'
 
 import AccessPointStatus from '../AccessPointStatus'
 
-const VanillaTorDetails = ({ measurement, render }) => {
-  const testKeys = measurement.test_keys
-  const isOK = testKeys.success === true
-  const isAnomaly = testKeys.success === false
-  const torVersion = testKeys.tor_version
-  const torLog = testKeys.tor_log
-  const torProgress = testKeys.tor_progress
-  const isFailure = testKeys.error !== null
-  const failure = testKeys.failure
-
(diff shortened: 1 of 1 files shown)


components/measurement/nettests/VanillaTor.js
```diff
@@ -10,16 +10,7 @@ import { MdTimelapse } from 'react-icons/md'
...
-  const testKeys = measurement.test_keys
-  const isOK = testKeys.success === true
-  const isAnomaly = testKeys.success === false
-  const torVersion = testKeys.tor_version
-  const torLog = testKeys.tor_log
-  const torProgress = testKeys.tor_progress
-  const isFailure = testKeys.error !== null
-  const failure = testKeys.failure
-
+const VanillaTorDetails = ({ measurement, render, isAnomaly, isFailure }) => {
```
majakomel, line 13: We only have 2 statuses for the VanillaTor screen - 'anomaly' and 'reachable' - probably we want to add 'failure' as well?

majakomel:
There are some other tests that are determining anomaly etc within the component (Signal, Psiphon..) - should we apply the same logic for them too?

components/measurement/nettests/VanillaTor.js
hellais: Yeah we should ideally use the `isFailure` flag as well and implement the failure screen in those cases. We probably could even make this into something general purpose to use in most measurement pages. It's basically just a matter of coloring the header differently and showing the failure screen.

hellais:
> There are some other tests that are determining anomaly etc within the component (Signal, Psiphon..) - should we apply the same logic for them too?

Yeah eventually we should do the same for those as well, however we ought to probably first check if the fastpath scoring is supporting everything that explorer supports, given that for some tests the explorer based anomaly detection was more feature rich compared to that of the fastpath.

Issue #875, opened 2023-10-17
 riseupvpn: relax presentation logic


bassosimone:
This issue is about changing the presentation logic of riseupvpn measurements to present them in a neutral way, like we do for ndt and dash measurements. Related backend issue https://github.com/ooni/backend/issues/745.

bassosimone:
I am going to transfer ownership of this issue to @majakomel

bassosimone:
I am going to transfer ownership of this issue to @majakomel

PR #877, opened 2023-10-18
fix(riseupvpn): render as an experimental nettest

bassosimone:
Closes https://github.com/ooni/explorer/issues/875


bassosimone:
Here's how the search renders:

Screenshot 2023-10-18 at 19 05 01

And here's how a measurement renders:

Screenshot 2023-10-18 at 19 05 42

majakomel:
This looks good! The only other place where RiseupVPN appears is here https://github.com/ooni/explorer/blob/master/components/search/FilterSidebar.js#L61

I believe it's safe to remove it here as well since those measurement are not going to be annotated?

PR #884, opened 2023-11-12
CustomTooltip: change link to measurements

### LINKED ISSUE #864: MAT: when the bars are grouped by week, the search links point to the wrong page
When I click on a bar with weekly granularity, the filter in the search is still for an individual day, while it should instead be filtering by the week.

Lanius-collaris:
If time grain isn't "day", use ( start + time grain ) as until

Resolves #864

### DIFF
--- a/components/aggregation/mat/CustomTooltip.js
+++ b/components/aggregation/mat/CustomTooltip.js
@@ -56,7 +56,12 @@ export const generateSearchQuery = (data, query) => {
         untilDateObj.setUTCDate(untilDateObj.getUTCDate() + 1)
         break
     }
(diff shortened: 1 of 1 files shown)


Lanius-collaris:
CC @hellais

majakomel (changes requested):
Hi @Lanius-collaris, thanks for your contribution! The code looks good, but I'd like to ask for a change in behaviour - currently time is included for all `until` filters, but it would preferably be only date when `time_grain` is day/week/month (hour is fine as is).

majakomel (approved):
Thank you for working on this, looks good!

PR #947, opened 2024-06-28
Replace styled-components with tailwind, upgrade ooni-components

majakomel:
This PR includes following changes:

- [x] Replace styled-components styling with Tailwind
- [x] Replace `<Button />` component with <button /> and use utility classes `btn` for styling
- [x] Replace `<Heading />` component with HTML Section Heading (styled via Tailwind base)
- [x] Remove `<Box />`, `<Flex />`, `<Container />`, `<Text />`, `<Link />`, `<Image />` and style them with Tailwind utility classes instead


components/Badge.js
```diff
@@ -1,35 +1,38 @@
...
+}
 
 const TestGroupBadge = ({ testName, ...props }) => {
   const { icon, groupName, color } = getTestMetadata(testName)
 
   return (
-    <Badge bg={color} color="white" {...props}>
-      <Flex sx={{ gap: 1 }} alignItems="center">
-        <Text>{groupName}</Text>
+    <Badge style={{ backgroundColor: color }} {...props}>
```
hellais, line 31: Q: why are we inlining the style instead of using the tailwind className

components/Layout.js
```diff
@@ -1,94 +1,59 @@
-import { Box, theme } from 'ooni-components'
 import PropTypes from 'prop-types'
-import React, { useMemo } from 'react'
-import { ThemeProvider, createGlobalStyle } from 'styled-components'
+import { useMemo } from 'react'
 
-import { StyledStickyNavBar, StyledStickySubMenu } from 'components/SharedStyledComponents'
-import { getDirection } from 'components/withIntl'
+import { StyledStickyNavBar } from 'components/SharedStyledComponents'
+// import { getDirection } from 'components/withIntl'
```
hellais, line 5: Maybe if we don't need this we can just remove the commented out code or mention why it's commented out

hellais:
I did a quick visual pass and left few minor comments.

All in all, when those are resolve this is good to go for me.

Thanks for putting this together and very glad to see this important and tedious work land 🥳

